PR #176, opened 2026-03-09
feat: OIDC

IamTaoChen:
### Summary
- Added OIDC authentication support using `kovah/laravel-socialite-oidc`
- Introduced per-provider `XXX_ENABLED` flags in `.env`:
  - `GOOGLE_ENABLED` (default: true if configured)
  - `GITHUB_ENABLED` (default: true if configured)
  - `OIDC_ENABLED` (default: false)
- In `social_login_buttons.blade.php`, each button is conditionally rendered based on its `ENABLED` flag + required config values
- OIDC button supports custom display name (`OIDC_DISPLAY_NAME`) and icon (`OIDC_ICON`)

### .env Example (added to .env.example)
```env
GOOGLE_ENABLED=true
GOOGLE_CLIENT_ID=
GOOGLE_CLIENT_SECRET=
GOOGLE_REDIRECT_URI=

GITHUB_ENABLED=true
GITHUB_CLIENT_ID=
GITHUB_CLIENT_SECRET=

# OIDC
# Optional: Enable or disable OIDC authentication (default: false)
# Only valid if OIDC_ISSUER, OIDC_CLIENT_ID, and OIDC_CLIENT_SECRET are set
OIDC_ENABLED=false
# OIDC issuer URL (required if OIDC is enabled)
OIDC_ISSUER=https://example.keycloak.com/realms/master
# OIDC client ID and secret (required if OIDC is enabled)
OIDC_CLIENT_ID=
OIDC_CLIENT_SECRET=
# Optional: OIDC scopes (default: "openid profile email")
OIDC_SCOPES=
# default is /auth/callback/oidc, but can be customized if needed
OIDC_REDIRECT_URI=/auth/callback/oidc
# Optional: Display name for the OIDC provider (default: "OIDC")
OIDC_DISPLAY_NAME=OIDC
# Optional: Icon URL for the OIDC provider (e.g., for login buttons)
OIDC_ICON=
# Optional: Enable JWT signature verification (default: false)
OIDC_VERIFY_JWT=false
# Optional: Provide a specific public key for JWT verification
#If not provided, the key will be fetched from the OIDC provider's JWKS endpoint
OIDC_JWT_PUBLIC_KEY=
```

***Developed with assistance from ChatGPT (code structure, conditionals, PR wording)***


mgarmash:
Any updates here? Can we proceed with the minimal OIDC version?

ManukMinasyan:
Pushed a fresh round of changes to this branch — merged latest `main`, regenerated `composer.lock`, and patched a security issue surfaced during review (the email-fallback in `CallbackController` could let a misconfigured/compromised IdP take over an existing account on email match alone — now requires `email_verified=true` from any non-Google/GitHub provider).

This is auth code touching every user, so I want to test it carefully before merging — at minimum exercising each of the 5 SSO providers (Keycloak, Okta, Azure, Authentik, Auth0) end-to-end against a real IdP, plus the email-collision path. Automated coverage is in place (parametrized tests for all 5 + 3 dedicated security-path tests, all green locally), but live IdP smoke tests are the gap.

If you have time to pull this branch and try it against your own setup, that would unblock things considerably:

```
gh pr checkout 176
composer install
php artisan migrate
# configure one of: KEYCLOAK_*, OKTA_*, AZURE_*, AUTHENTIK_*, AUTH0_* in .env (set *_ENABLED=true)
php artisan test --filter=SocialiteLoginTest
# then open the app and try the SSO button
```

A quick "works for me on $provider" report from a real deployment is worth more than another pass from me. Happy to keep iterating in parallel.

mgarmash:
Tested, works fine for me with Authentik. 
Just a few notes. It would be helpful to disable registration while SSO mode is on. And remove login and password logic from login screen. Besides that, we should think about fixing some workflows with inviting people by email and adding them directly. I think it should behave according to selected mode. 
But anyway it's already useable. Will try for real workflow nearest workday.

Issue #24, opened 2025-08-19
Filament 4 ?


anielka41:
Hello, when are you planning to upgrade to filament4?
I see that you already have the filament4 structure in your files, but I don't know when you will finish it.

ManukMinasyan:
@anielka41 I am actively working to upgrade to V4. This effort involves not only this repository but also the Flowforge Kanban board, Custom Fields, and one additional new plugin. I plan to complete this work within two weeks.

Issue #17, opened 2025-07-04
Fix Missing 'code' Parameter in Google OAuth2 Token Request


ManukMinasyan:
`A GuzzleHttp\Exception\ClientException` is thrown when making a POST request to https://www.googleapis.com/oauth2/v4/token, resulting in a 400 Bad Request response with the error message: `"Missing required parameter: code"`. This indicates that the required code parameter is not being included in the token request payload.

ManukMinasyan:
https://github.com/Relaticle/relaticle/commit/66770f36f227072940b8fddcf69d33d9a14ffca0 - Fixed

PR #237, opened 2026-04-23
Feat/email calendar integration

### LINKED ISSUE #91: Email & Calendar Integration
# Email & Calendar Integration

Full email and calendar integration for Relaticle — sync, send, view, and manage communications directly within the CRM. Privacy-first by design: all emails are private by default, with granular sharing controls.

---

## Sub-Issues

- [ ] Gmail OAuth 2.0 connection flow
- [ ] Microsoft 365 / Outlook OAuth connection flow
- [ ] Sync inbox and sent folders (historical + ongoing)
- [ ] Auto-create person records from email participants (configurable: all contacts / bidirectional only / none)
- [ ] Auto-create company records by domain matching (skip public domains like gmail.com, yahoo.com, etc.)
- [ ] Background sync worker (queue-based, periodic polling)
- [ ] Account connection management UI (connect, disconnect, re-authenticate)
- [ ] Multi-account support (user can connect multiple mailboxes)

- [ ] **Default privacy:** Emails visible only to the syncing user
- [ ] **3 sharing tiers:**
  - Metadata only (participants + timestamps)
  - Subject line + metadata
  - Full access (body, attachments, everything)
- [ ] Workspace-level default sharing setting (admin configurable)
- [ ] Per-email manual sharing (share individual emails with team or specific people)
- [ ] Per-record sharing (share all your emails on a specific record)
- [ ] **Auto-hide internal emails** (between workspace members — automatic)
- [ ] **Blocklist:** Hide emails from specific addresses or domains (per user)
- [ ] **Protected recipients** (admin-level): Hide all communications for sensitive contacts/companies across entire workspace
- [ ] Access request flow: teammate requests access → owner approves/denies via notification
- [ ] Sharing permissions UI in email detail view

- [ ] Emails tab on person records (matched by email address)
- [ ] Emails tab on company records (matched by email domain)
- [ ] Emails tab on deal records (linked via associated contacts/companies)
- [ ] Activity timeline integration (emails + meetings in chronological feed)
- [ ] Email detail view (subject, body, attachments, participants, timestamps)
- [ ] Respect privacy controls — show metadata/subject/full based on sharing tier
- [ ] AI-generated email thread summaries (extend existing Claude API integration)
- [ ] Auto-labels: categorize emails (Scheduling, Marketing, Invoices, Support, etc.)
- [ ] Search within emails on a record
- [ ] Communication intelligence attributes: "Last interaction," "Days since last email"

- [ ] Compose email from record page (person, company, deal)
- [ ] Quick compose via keyboard shortcut
- [ ] To, CC, BCC fields with contact autocomplete
- [ ] Rich text editor (formatting, inline images)
- [ ] File attachments
- [ ] Multiple signatures per account (create, edit, set default)
- [ ] **Email templates** with variable personalization (`{first_name}`, `{company}`, custom field values)
- [ ] Template management UI (create, edit, organize)
- [ ] Reply / Reply All / Forward from email detail view
- [ ] Sent emails auto-linked to relevant records
- [ ] Mass send: batch emails to multiple recipients (each gets individual thread)

- [ ] Outbox UI — view queued, sending, sent, failed emails
- [ ] Rate limiting per mailbox (e.g., 12/hr, 200/day) to protect deliverability
- [ ] Sending priority: single sends → mass sends
- [ ] Preview queued emails before they send
- [ ] Cancel or edit emails still in queue
- [ ] Retry failed emails with error details
- [ ] Max scheduled email limit per user

- [ ] Google Calendar OAuth connection
- [ ] Outlook Calendar OAuth connection
- [ ] Auto-link meetings to person/company records (by attendee email/domain)
- [ ] Meetings tab on records + activity timeline
- [ ] Meeting detail view (title, time, attendees, description)
- [ ] Private/declined events excluded from sync
- [ ] Manual meeting-to-record linking
- [ ] Calendar connection management UI

- [ ] Unique workspace inbound address (`workspace@relaticle.email`)
- [ ] Accept forwarded and BCC'd emails from registered users only
- [ ] Respect auto-record-creation settings
- [ ] Forwarded email attachments preserved
- [ ] Link forwarded emails to matching records
- [ ] Good for providers not supported by OAuth (alternative sync path)

---

## Priority Order

| Phase | Sub-Issues | Goal |
|-------|-----------|------|
| **Phase 1** | 1 (Sync) + 2 (Privacy) + 3 (View) | Foundation — emails flow in, privacy-first, visible on records |
| **Phase 2** | 4 (Send) + 5 (Outbox) | Users can send from CRM with deliverability protection |
| **Phase 3** | 6 (Calendar) + 7 (Forwarding) | Full communication hub |

## Design Principles

- **Privacy-first:** All emails private by default. Users opt-in to sharing, never opt-out.
- **Auto-linking:** Emails automatically appear on the right records — zero manual work.
- **Agent-native:** MCP tools can read email context (respecting privacy) for AI-powered insights.
- **No IMAP/SMTP:** OAuth-only (Gmail + Outlook) for security and reliability.

AsmitNepali:
## Summary

Adds end-to-end email and calendar integration plus a standalone Activity Log plugin to the CRM. Covers 511 commits, 457 files, and broad test coverage for the new flows.

## Email Integration (`packages/EmailIntegration`)

- **Inbox and detail view.** Filament page with folder tabs (All / Inbox / Sent), search, unread counts, and a full detail pane with Reply, Reply All and Forward.
- **Compose, reply, forward, scheduled send.** Rich editor with templates, signatures, merge tags, attachments, and a 30-second undo window.
- **Privacy tiers and sharing.** Per-email privacy (`PRIVATE`, `METADATA_ONLY`, `SUBJECT`, `FULL`), per-teammate shares, and a bulk "Share my emails on this record" action.
- **Access requests.** Request, approve, deny and revoke flow with notifications and a dedicated settings page.
- **Relation managers.** Emails tab on People, Company and Opportunity with visibility-aware columns, sharing actions, and AI thread summaries.
- **Per-record Emails page.** Full inbox-style page scoped to a CRM record.
- **Blocklist, signatures, templates, outbox.** Management pages and services.
- **Reply and forward inside the relation-manager view modal.** Wires up the missing `reply-email` listener on `BaseEmailsRelationManager` so the buttons inside the email view modal mount the reply action.

## Calendar Integration

- Google Calendar OAuth (redirect and callback, state-aware), a per-account toggle, and a manual "Sync now" action.
- Meetings resource and `BaseMeetingsRelationManager` (People / Company / Opportunity).
- Incremental sync job, `StoreMeetingAction` with filter and upsert, link and unlink actions, meeting observer.

## Activity Log Plugin (`Plugins/ActivityLog`)

- Pluginised timeline with `TimelineBuilder`, sources (activity log, related models, custom events), renderer registry, infolist entry, relation manager, and a Livewire component with infinite scroll.
- Replaces the in-app timeline with the plugin on People, Company and Opportunity views.

## Setup

The integration sits behind a feature flag and needs its own Google OAuth client. It does not reuse the credentials behind Google sign in.

### 1. Environment

```dotenv
RELATICLE_FEATURE_EMAIL_INTEGRATION=true

GMAIL_CLIENT_ID=
GMAIL_CLIENT_SECRET=
GMAIL_REDIRECT_URI="https://your-app-host/email-accounts/callback/gmail"
```

These map to `services.gmail` in `config/services.php`. The callback route sits at the application root, not under the panel path.

### 2. Google Cloud project

Create a **dedicated project** for the email integration. Do not add these scopes to the project backing Google sign in. Branding, scopes, publishing status and verification state are all project-level, so Gmail scopes would drag the sign-in app into restricted-scope review it does not otherwise need. Google separately requires one project per deployment tier.

In that project:

1. Enable the **Gmail API** and the **Google Calendar API**.
2. Configure the consent screen as **External**, publishing status **Testing** during development, and add every mailbox you plan to connect as a test user.
3. Register these scopes under Data Access:
   - `https://www.googleapis.com/auth/gmail.readonly`
   - `https://www.googleapis.com/auth/gmail.send`
   - `https://www.googleapis.com/auth/calendar.readonly`
4. Create an OAuth client of type **Web application**, and add the value of `GMAIL_REDIRECT_URI` as an authorized redirect URI. It must match character for character.
5. Copy the client id and secret into the environment variables above.

### 3. Local development

Google rejects redirect URIs on the `.test` TLD, so a local host needs a public https URL that forwards to it. Wrap the local callback in a public forwarder such as `fwd.host` and register the wrapped form in both the console and `GMAIL_REDIRECT_URI`:

```
https://fwd.host/https://your-app-host.test/email-accounts/callback/gmail
```

The forwarder sees the authorization code in transit, so use this only with a development OAuth client, never in production.

After changing any of these values:

```bash
php artisan config:clear
php artisan horizon:terminate   # workers hold the old credentials until restarted
```

The queue worker matters here. Connecting an account dispatches the history import and initial sync jobs, and token refresh reads the same credentials at job runtime.

### 4. Syncing

Incremental sync and outbox dispatch run on the scheduler. To drive them by hand while testing:

```bash
php artisan email:incremental-sync
php artisan calendar:incremental-sync
php artisan email:dispatch-outbox
```

### Known constraints

- A project in **Testing** status is capped at 100 test users, and authorizations plus refresh tokens **expire seven days after consent**. A connected mailbox needs re-consent weekly during development. That is Google behavior, not a defect in the sync code.
- Connecting grants mail scopes only. The calendar scope is requested separately when calendar sync is enabled on an account, following Google's guidance to request scopes incrementally.
- `gmail.readonly` is a **restricted** scope. Production use requires OAuth verification plus an annual third-party security assessment. That review runs for weeks and needs to land before the flag is enabled broadly.

## Test Plan

- [ ] Connect a Google account, verify the inbox populates and mark-as-read works
- [ ] Compose, schedule, undo, reply, reply all, forward
- [ ] Exercise privacy tiers: set per email, request access, then approve, deny and revoke
- [ ] Open the Emails tab on People, Company and Opportunity. Reply and forward fire from the view modal, and the sharing, summarize and request-access row actions still work
- [ ] Enable calendar sync on the account, run a manual sync, link and unlink a meeting, confirm the observer and relation manager update
- [ ] Activity Log timeline shows on People, Company and Opportunity with correct entries and pagination
- [ ] `vendor/bin/pint --format agent`, `vendor/bin/phpstan analyse`, `composer test:type-coverage`, `php artisan test` all pass


ManukMinasyan (changes requested):
## Deep Review

Large feature PR (282 files, +22,729/−433, 117 commits). Implementation is generally well-structured with policies, factories, and substantial test coverage — but the privacy/sharing layer has multiple cross-tenant authorization bypasses, the inline email body renderer is exploitable for stored XSS, and several project conventions are violated.

---

### Critical

| # | Category | File:Line | Issue | Suggestion |
|---|----------|-----------|-------|------------|
| 1 | Security | `resources/views/filament/emails/email-view.blade.php:193-198` and `email-thread.blade.php:170-176` | Email HTML rendered in iframe with `sandbox="allow-same-origin allow-popups"`. Same-origin sandbox = scripts inside the iframe can read `document.cookie`, fire same-origin XHRs against `/admin/...`, and fully impersonate the user. The `preg_replace('/<script\b[^>]*>.*?<\/script>/is', '', ...)` filter is trivially bypassable (`<svg onload>`, ``, `<iframe srcdoc>`, `javascript:` URLs, malformed tags). One inbound email = session takeover. | Drop `allow-same-origin` (single most important fix). Or serve email body from a dedicated sandbox subdomain via a controller. Add a real HTML purifier (`mews/purifier` / `ezyang/htmlpurifier`) with strict tag/attribute allowlist. |
| 2 | Security | `app/Filament/Pages/EmailInboxPage.php:394-415` (and same pattern in `app/Filament/RelationManagers/BaseEmailsRelationManager.php`) | `manageSharing` action does `Email::query()->whereKey($arguments['emailId'])->first()` with no `team_id`, ownership, or policy check. Any authenticated user who knows/guesses an email ULID can change its `privacy_tier`, wipe existing shares, and grant `User::findOrFail($share['shared_with'])` (also unscoped) access to another tenant's emails. | Replace with `$this->authorize('share', $email)` against `EmailPolicy`. Require `team_id === current_team_id` and `user_id === auth()->id()`. Validate `shared_with` belongs to the same team. |
| 3 | Security | `packages/EmailIntegration/src/Actions/SendEmailAction.php:48` | `ConnectedAccount::query()->findOrFail($data['connected_account_id'])` with no ownership check. `connected_account_id` flows from a Livewire form payload (client-controlled). An authenticated user can send mail through another tenant's Gmail account (impersonation, sender-reputation damage). | Inside the action, assert `$account->user_id === auth()->id()` and team membership, or accept a resolved `ConnectedAccount` model from the call site after authorization. |
| 4 | Security | `packages/EmailIntegration/src/Filament/Pages/EmailSignaturesPage.php:99-115` (`create`) and `:118-166` (`edit`) | `createSignatureAction` and `editSignatureAction` both look up the connected account / signature with `whereKey(...)->firstOrFail()` and zero ownership scoping. Combined with raw `{!! $signature->content_html !!}` rendering on the signatures page, an attacker can edit a victim's signature into an XSS payload that fires the next time the victim opens settings. (Only `deleteSignatureAction:178` is correctly scoped.) | Scope every lookup by `whereHas('connectedAccount', fn ($q) => $q->where('user_id', auth()->id())->where('team_id', filament()->getTenant()?->getKey()))`. Sandbox the rendered signature HTML. |
| 5 | Security / Logic | `packages/EmailIntegration/src/Models/Scopes/VisibleEmailScope.php:27-31` | Operator precedence bug. The closure resolves to `(is_internal=false AND privacy_tier!=PRIVATE) OR shares-exists`. Any email a viewer was once shared on becomes visible *even if it's PRIVATE or internal*. Also no `team_id` floor — any future caller of the scope without an explicit team filter leaks across tenants. | Wrap the OR branch in a nested closure so `shares-exists` is grouped with `(internal AND !PRIVATE)`. Add `where('team_id', $this->viewer->current_team_id)` as a hard floor inside the scope. |
| 6 | Migrations | `database/migrations/2026_03_21_054952_create_email_signatures_table.php` (entire file) | `email_signatures` is missing `team_id` column, FK, and index. Every other email table in the PR has `team_id`; this is the only one. Multi-tenant queries can't scope correctly without joining through `connected_accounts`. | Add `$table->foreignUlid('team_id')->constrained()->cascadeOnDelete();` and an index on `(team_id, user_id)`. |
| 7 | Convention | 20 migration files in `database/migrations/2026_03_*.php` and `2026_03_29_045319_add_email_metrics_to_opportunities_table.php` | All have `down()` methods. Project rule (`CLAUDE.md`): "Migrations must only have `up()` methods — do not write `down()` methods." Affected files include `2026_03_21_050414_create_email_accounts_table`, `..._050430_create_emails_table`, `..._050442_create_email_participants_table`, `..._050455_create_email_bodies_table`, `..._050506_create_email_attachments_table`, `..._050520_create_email_shares_table`, `..._050535_create_email_blocklists_table`, `..._050549_create_protected_recipients_table`, `..._054333_create_emailables`, `..._054840_create_email_threads_table`, `..._054856_create_connected_account_syncs_table`, `..._054912_create_email_access_requests_table`, `..._054922_create_public_email_domains_table`, `..._054937_create_email_templates_table`, `..._054952_create_email_signatures_table`, `..._055334_add_default_email_sharing_tier_to_teams`, `..._055435_add_email_intelligence_to_people_and_companies`, `..._055500_create_email_labels_table`, `2026_03_25_000001_add_default_email_sharing_tier_to_users_table`. | Strip every `down()` block. The 2026_04_20_* migrations correctly omit them. |
| 8 | Logic | `packages/EmailIntegration/src/Jobs/InitialEmailSyncJob.php:46` | `Config::integer('email-integratdion.sync.initial_days', 90)` — typo (`integratdion`). Config never resolves; fallback `90` is always used regardless of `config/email-integration.php`. | Fix to `email-integration.sync.initial_days`. |
| 9 | Logic | `packages/EmailIntegration/src/Actions/StoreEmailAction.php:70 & :87` | `LinkEmailAction::execute($email)` is called twice in the same transaction. Doubles the auto-link work and can double-bump `email_count` columns updated via `DB::raw('email_count + 1')` in `LinkEmailAction`. | Remove the second invocation; if the second call is for `is_internal` reclassification, refactor `LinkEmailAction` to be idempotent or split the responsibility. |
| 10 | Dependencies | `composer.json:42` | `"socialiteproviders/microsoft-azure": "*"` — wildcard constraint. No upper bound; any future major release silently lands on `composer update`. | Pin to `^5.2`. |
| 11 | Dependencies | `composer.json:36` | `"relaticle/activity-log": "1.x-dev"` — dev-branch reference (no tagged release) combined with `minimum-stability: beta`. Any push to the package's `1.x` branch can change behavior on the next `composer update`. | Tag `1.0.0` of `relaticle/activity-log` and pin to `^1.0`. |

---

### Warnings

| # | Category | File:Line | Issue | Suggestion |
|---|----------|-----------|-------|------------|
| 12 | Security | `packages/EmailIntegration/src/Actions/ApproveEmailAccessRequestAction.php:17-38` (and `DenyEmailAccessRequestAction.php`) | Approve/deny/share grants happen with no actor argument — gating lives only in the calling Filament page. Defense-in-depth missing for a security-critical primitive. | Pass the acting `User` into `execute()` and assert `$accessRequest->owner_id === $actor->getKey()` inside the action. |
| 13 | Security | `packages/EmailIntegration/src/Filament/Pages/EmailAccountsPage.php:88-106, 157-203` | `editSettingsAction::fillForm`, `syncCalendarAction::action`, `syncCalendarNowAction::action` use bare `ConnectedAccount::query()->findOrFail(...)` — only the `update()` itself is scoped. Cross-tenant settings read, foreign sync toggle, foreign sync-job dispatch are all possible. | Add a private `findOwnedAccount(string $id)` helper that scopes by `user_id` + `team_id` and use it everywhere. |
| 14 | Security | `app/Http/Controllers/EmailAttachmentController.php:50-56` | Streams attachments back with provider-supplied `Content-Type` and no `X-Content-Type-Options: nosniff`. A sender can craft `text/html` attachments that Chrome will render inline at the app origin → session theft. | Force `Content-Disposition: attachment`, add `X-Content-Type-Options: nosniff`, and either whitelist mime types or always serve as `application/octet-stream`. |
| 15 | Logic | `app/Filament/Pages/EmailInboxPage.php:806` | `buildSendData()` hardcodes `'privacy_tier' => EmailPrivacyTier::FULL` for every outbound email, ignoring the team/user default in `PrivacyService::defaultTierForUser()`. Privacy regression that silently contradicts the PR's stated model. | Read the workspace/user default and let the compose form override. |
| 16 | Logic | `packages/EmailIntegration/src/Actions/CancelQueuedEmailAction.php:20` + `app/Jobs/SendEmailJob.php` | Cancel only allows `status === QUEUED`. Dispatcher flips to `SENDING` once due. If the user clicks "Undo" within the same second the dispatcher fires, the cancel throws "Too late" even though Gmail hasn't been called yet. | Allow cancel during `SENDING` while `provider_message_id IS NULL`, or guard the dispatcher with a small safety margin past `scheduled_for`. |
| 17 | Logic | `app/Jobs/SendEmailJob.php:71-87` | `failed()` only persists `$exception->getMessage()` — class and trace are lost. Combined with `$tries=3`, debugging deliverability from the outbox alone is hard. | `Log::error('SendEmailJob failed', ['exception' => $exception])` or store class+message together. |
| 18 | Convention (Action-pattern) | `app/Filament/Pages/EmailInboxPage.php:404-415, 483-489`; `app/Filament/RelationManagers/BaseEmailsRelationManager.php:253-262, 299-307`; `app/Livewire/App/Email/UserEmailPrivacySettings.php:109-131`; `packages/EmailIntegration/src/Filament/Pages/EmailPrivacySettingsPage.php:66-92`; `packages/EmailIntegration/src/Filament/Pages/EmailAccountsPage.php:142-225` | Project rule (`CLAUDE.md`): "All write operations must go through action classes." These all create/update/delete Eloquent models inline in pages, relation managers, or Livewire components. | Extract each into a dedicated `app/Actions/...` (or package action) class. |
| 19 | Dependencies / Operations | `config/horizon.php` (new supervisors) vs `packages/EmailIntegration/src/Jobs/IncrementalEmailSyncJob.php`, `IncrementalCalendarSyncJob.php`, `app/Jobs/ClassifyEmailJob.php` | New supervisors `supervisor-emails-priority` (queue `emails-priority`) and `supervisor-emails-bulk` (`emails-bulk`) only handle outbound `SendEmailJob`. The three sync/classify jobs set no queue and land on `default`. Recurring sync (every N minutes per active account) can starve the default supervisor. | Either dispatch sync/classify jobs onto a dedicated `emails-sync` queue with its own supervisor, or document the intent. |
| 20 | Dependencies | `packages/EmailIntegration/` (no `composer.json`) | The package declares no `composer.json`, so `google/apiclient` and `socialiteproviders/microsoft-azure` are runtime deps of the package but only declared in the host `composer.json`. Consistent with other in-tree packages, but fragile. | Optional but recommended: add `packages/EmailIntegration/composer.json` declaring runtime deps explicitly. |
| 21 | Migrations | `database/migrations/2026_03_21_054333_create_emailables.php:13-22` | No unique index on `(email_id, emailable_type, emailable_id)`. Auto-link logic on retry can create duplicate links. (Sister table `meetingables` in `2026_04_20_000004` does this correctly.) | Add `$table->unique(['email_id', 'emailable_type', 'emailable_id'], 'emailables_unique_link');`. |
| 22 | Migrations | `database/migrations/2026_03_21_050430_create_emails_table.php:25` and `2026_03_21_054840_create_email_threads_table.php` and `..._054937_create_email_templates_table.php` | `subject` is `string` (255). Real subjects (newsletters, deep `Re: Re:` chains) routinely exceed 255 bytes; Gmail sync will throw. | Use `text` for `subject` columns. |
| 23 | Logic / Pollution | `packages/EmailIntegration/src/Actions/LinkEmailAction.php:43-46` | `whereRaw('json_value::text like ?', ["%{$domain}%"])` matches substring collisions: `acme.co` matches `acme.com.au`. Also LIKE wildcards `_` and `%` from sender input are not escaped. | Use `whereJsonContains('json_value', $domain)` (the same file already does this on line 66 for the email lookup). |
| 24 | Convention | `packages/EmailIntegration/src/Observers/ConnectedAccountObserver.php:15`; `packages/EmailIntegration/src/Filament/Pages/EmailAccountsPage.php:67`; `routes/web.php:63` | TODO/FIXME comments left in: "TODO::remove this logic from observer", "TODO::Implement after azure setup", "TODO::Remove after check all features." | Resolve before merge or convert into tracked issues with a code reference. |
| 25 | Code Quality | `app/Filament/Pages/EmailPrivacySettingsPage.php` AND `packages/EmailIntegration/src/Filament/Pages/EmailPrivacySettingsPage.php` | Two classes with the same name and overlapping intent (both register under nav group "Emails"). Different slugs (`email-privacy` vs `settings/email-privacy`) but they will drift. | Consolidate into one. |

---

### Info

| # | Category | File:Line | Issue | Suggestion |
|---|----------|-----------|-------|------------|
| 26 | Security | `.env.testing` | `APP_KEY` is committed (testing-only). | Move to CI-injected env. |
| 27 | Code Quality | `vite.config.js`, `.gitignore` (`/packages/filament` ignore), `package-lock.json` rename `fierce-shark→relaticle` | Cosmetic / unrelated to feature. | Drop the `.gitignore` entry if it's a local scratch dir; otherwise fine. |
| 28 | Migrations | `database/migrations/2026_03_21_050414_create_email_accounts_table.php` | Filename says `email_accounts` but it actually creates the `connected_accounts` table. | Rename for clarity (low priority). |
| 29 | Migrations | `database/migrations/2026_03_21_050520_create_email_shares_table.php`; `..._050549_create_protected_recipients_table.php` | Missing or unindexed `team_id` (shares has none; protected_recipients has the column but no index). Cross-team scoping relies on app-layer joins. | Add `team_id` to `email_shares` (and an index) for cheap scoping; add `(team_id, type, value)` index on `protected_recipients`. |
| 30 | OAuth callback | `packages/EmailIntegration/src/Controllers/CallbackController.php:23-66` | Socialite's default state validation is in place. No try/catch around `$driver->user()` — `InvalidStateException` shows a stack trace if `APP_DEBUG=true`. | Wrap in try/catch and redirect with a friendly notification. |
| 31 | Scope | `app/Observers/CustomFieldValueObserver.php` (+100 lines, new) and changes to `Note.php`, `Task.php`, `CustomFieldValue.php`, `ApplyTenantScopes.php` | Activity-log integration tangentially related to the email PR. Defensible if shipping the activity-log plugin together, but bundling makes review noisy and rollback harder. | Consider splitting the activity-log changes into a separate PR. |

---

### Things that look clean

- OAuth tokens encrypted at rest via Eloquent `encrypted` cast on `ConnectedAccount`.
- Socialite state validation in place.
- `EmailAttachmentController` properly checks `team_id` and `viewBody` policy (only response headers need fixing — see #14).
- No `dd()`/`dump()`/`ray()` left behind.
- No raw SQL injection vectors in new code (LIKE-wildcard issue is data quality, not injection).
- Foreign-key cascade behavior on email child tables is correct.
- Migrations don't add NOT-NULL columns without defaults; existing data is safe.
- PostgreSQL-only — no driver branching anywhere in new code.
- `bootstrap/providers.php` registration is correct; package providers spelled right.
- `email:dispatch-outbox` IS scheduled (`bootstrap/app.php:95`).
- `CLAUDE.md` `=== .ai/v4 rules ===` deletion is correct (project upgraded to Filament 5).

---

### Verdict

**Request changes.** The PR delivers a substantial and largely well-engineered feature, but cannot ship as-is. The five Critical security findings (#1–#5) plus the privacy-scope bug (#6 missing `team_id`) are exploitable cross-tenant data-integrity and code-execution holes — especially the iframe XSS (#1) and the manage-sharing bypass (#2), which together let any authenticated user take over any other user's session via a crafted email or grant themselves access to any email in the system. Critical convention/data findings #7–#11 (down() methods, missing `team_id` on signatures, config typo, version pinning) are quick fixes that should land in the same revision. Warnings should be addressed in the same revision; Info items can be deferred.

ManukMinasyan:
# Business Review — PR #237 "Email & Calendar Integration"

**Reviewed commit:** `c379331a6c` (HEAD of `feat/email-calendar-integration`)
**Scope:** 335 files, +26,945 / −13, 28 migrations, 117 commits. Feature lives in `packages/EmailIntegration` + app wiring.
**Mode:** Browser verification against the PR preview (`app.relaticle-pr-237.test`) + targeted code verification of the prior deep-review criticals.

---

## Verdict: ⚠️ AI-NEEDS-HUMAN — strong implementation, **not shippable to 100k as-is**

The feature is well built and the **11 security criticals from the prior `CHANGES_REQUESTED` deep review are all remediated at HEAD** (code-verified; the headline XSS fix confirmed **live** in the browser). This is far better than the surface signals (`CONFLICTING` + `CHANGES_REQUESTED`) suggest.

But it **cannot go to a 100k production deploy right now**. Hard blockers:
1. **Does not merge into `main`** (conflict; needs rebase).
2. **GitHub review state is still `CHANGES_REQUESTED`** (must be re-evaluated/dismissed by the reviewer).
3. **Core value — real Gmail/Calendar sync + send — was never validated end-to-end** (not possible in this environment; see boundaries).
4. **Google OAuth restricted-scope verification** is a launch prerequisite.
5. **No gradual rollout** — the feature flag is a global on/off defaulting to ON.

I did **not** merge to main, push, or deploy. Those need explicit human sign-off (details below).

---

## ✅ Security criticals — verified FIXED at HEAD

The prior deep review (by ManukMinasyan) flagged 11 criticals against an earlier commit (282 files). All are remediated at current HEAD:

| # | Issue | Status at HEAD | Evidence |
|---|-------|----------------|----------|
| 1 | Stored XSS → session takeover (iframe `allow-same-origin` + bypassable regex) | **FIXED + LIVE-CONFIRMED** | `sandbox="allow-popups allow-popups-to-escape-sandbox"` (no same-origin) + `App\Support\EmailHtmlSanitizer` (Symfony allowlist). Live test: seeded `<script>`/`onerror`/`onload`/`javascript:` payloads — all stripped, `window.__xss` stayed `undefined`. |
| 2 | `manageSharing` IDOR (rewrite any email's privacy/shares by ULID) | **FIXED** | `resolveTeamEmail()` → `forTeam()` + `$user->can($ability,$email)`; write via `UpdateEmailSharingAction`; teammate options scoped to `current_team_id`. |
| 3 | Send mail through another tenant's Gmail | **FIXED** | `ConnectedAccount::query()->ownedBy($user,$user->currentTeam)->whereKey(...)->firstOrFail()`. |
| 4 | Signature IDOR + XSS (`{!! content_html !!}`) | **FIXED** | Rendered via `HtmlSanitizerService::sanitizeRichText()`; lookups scoped. |
| 5 | `VisibleEmailScope` precedence + missing team floor | **FIXED** | `->where('team_id', viewer->current_team_id)` floor added; share/public-gate OR correctly nested. |
| 6 | `email_signatures` missing `team_id` | **FIXED** | `$table->teams()` + `index(['team_id','connected_account_id'])`. |
| 7 | Migrations with `down()` (convention) | **FIXED** | Zero `down()` methods remain in PR migrations. |
| 8 | Config typo `email-integratdion` (sync days) | **FIXED** | Now `email-integration.sync.initial_days`. |
| 9 | `LinkEmailAction` called twice (double-count) | **FIXED** | Single call in `StoreEmailAction`. |
| 10 | `socialiteproviders/microsoft-azure: "*"` | **FIXED** | Pinned `^5.2`. |
| 11 | `relaticle/activity-log: "1.x-dev"` | **FIXED** | Pinned `^1.0`. |

Warnings also fixed: **#12** access-request actions take `User $actor`; **#13** account actions use `findOwnedAccountOrFail`/`ownedBy`; **#14** attachment controller forces `octet-stream` + `nosniff` + `CSP: default-src 'none'; sandbox` + team/policy checks; **#15** compose uses `defaultPrivacyTier()` (not hardcoded FULL); **#16** undo cancel allowed during `SENDING` while `provider_message_id IS NULL` (row-locked); **#19** sync/classify jobs on dedicated `emails-sync` queue + `supervisor-emails-sync`; **#21** `emailables` unique link index; **#22** `subject` columns now `text`.

## ✅ Functional (browser, what was testable)

- Login, tenant dashboard ✅
- Inbox renders (All/Inbox/Sent tabs, empty + populated states) ✅, no console errors
- Accounts page: "Connect Gmail" present, Outlook hidden ✅
- OAuth initiation → Google consent with correct scopes (`gmail.readonly`, `gmail.send`) + CSRF `state` ✅
- Email detail renders sanitized body safely in opaque-origin iframe ✅
- Sharing modal: 4 privacy tiers + per-teammate shares, team-scoped ✅
- Pages render clean: email-privacy, email-signatures, email-templates, outbox, access-requests ✅
- Meetings page errored (`relation … does not exist`) — **environment artifact** (meetings table couldn't be migrated in the shared dev DB; see below), **not a code bug**.

## ⛔ Not testable here (honest boundaries)

- **Real Gmail/Outlook connect → sync → send:** OAuth round-trip can't complete into the preview (`redirect_uri` resolves to `app.relaticle.test` via `fwd.host`; `GOOGLE_REDIRECT` empty). Microsoft/Azure unconfigured. **The feature's core value is unverified end-to-end.**
- **Calendar/meetings:** needs Google Calendar OAuth + the meetings table.
- **"Stress testing real integrations" at 100k:** not feasible and **inadvisable** — would burn Gmail/Calendar API quota, risk sender-reputation/deliverability, and likely violate Google API ToS. This needs a controlled staging mailbox + a deliberate load plan, not ad-hoc stress against live providers.

---

## 🚧 Release blockers for a 100k deploy

1. **Rebase/merge `main`.** PR is `CONFLICTING` (258 commits behind). Conflicts in **2 non-feature files**: `app/Filament/Resources/OpportunityResource.php`, `tests/Arch/ArchTest.php`. Small, but must be resolved + re-verified.
2. **Resolve `CHANGES_REQUESTED`.** The blocking review must be re-reviewed/dismissed now that criticals are fixed.
3. **Validate real integrations** in a controlled env (connect a staging Gmail, sync, send, undo, reply/forward, calendar sync, link/unlink meeting).
4. **Google OAuth verification** for restricted scopes (CASA) — required before 100k external users.
5. **Rollout strategy.** Feature flag is global on/off (`config('relaticle.features.email_integration', true)`), defaulting ON — no per-tenant/percentage rollout. Recommend deploying flag-OFF then ramping, or implement per-team Pennant rollout.
6. **Migration deploy plan.** 28 migrations; dry-run on a prod-like dataset + rollback plan. **Coordinate the `activity_log` migration with branch `feat/activity-log`** — both introduce `create_activity_log_table` at different timestamps; whichever merges second will collide. (`origin/main` currently has neither, so #237 alone is safe.)

## Lower-priority / not re-verified
Prior-review items not individually re-checked: #17 (job `failed()` logging), #18 (write-in-page convention), #20 (package `composer.json`), #23 (LIKE escaping in `LinkEmailAction`), #24 (TODOs), #25 (dup `EmailPrivacySettingsPage` class), #26 (`.env.testing` APP_KEY), #28–31 (filename/indexes/OAuth try-catch/scope). Mostly minor/convention; none are 100k blockers on their own.

---

## Environment notes (for reproducers)
- Dev DB is **shared** across all Herd clones (`relaticle_app`). The `activity_log` migration "failure" during setup was this branch's migration colliding with `feat/activity-log`'s already-run one — a shared-DB artifact, not a #237 defect.
- Preview's `APP_PANEL_DOMAIN` was `app.relaticle.test` (a different checkout). Temporarily pointed it at `app.relaticle-pr-237.test` for testing; reverted after.
- Test data: connected account + emails prefixed `br-237-` (incl. one sanitization-test email with a neutralized XSS body). Left in place.

AsmitNepali:
**Merged #386 → this branch: per-viewer email read state**

Fixes a bug in this PR's email integration. Read state lived in single global `emails.read_at` column owned by the email's connector. Unread badge counts per-viewer (own + privacy-tier-shared emails), but `MarkEmailAsReadAction` gated to `user_id = owner` — so teammate opening a *shared* email = no-op, unread count never dropped.

Fix moves read state to per-`(email, user)` `email_reads` table (matches Front/Missive/HubSpot shared-inbox model):
- New `email_reads` table; `emails.read_at` removed (edited original create-table migration since branch unmerged).
- `EmailRead` model + `Email::reads()`, scopes `withReadStateFor($userId)` / `unreadFor($userId)`.
- `MarkEmailAsReadAction` upserts read row for acting user, guarded by `VisibleEmailScope` (no forging reads on invisible emails).
- Inbox page + per-record page use `unreadFor()`; list keys `is_read` off per-viewer flag.
- Owner provider-sync paths write/delete owner's read row only — teammate read state untouched by Gmail/Graph sync.

New `EmailUnreadCountPerViewerTest` + sync-job coverage. pint/rector/phpstan clean, type-coverage 100%.

AsmitNepali:
**Fix: duplicate auto-created companies from the same domain**

Email/calendar sync matches an incoming sender domain to a `Company` and auto-creates one when there's no match. The match and the create were separated by a gap, so two `StoreEmailJob`/calendar workers processing the first message from a brand-new domain in parallel could both miss the match and both create a company → duplicate companies for the same domain.

**What changed** (merged via #389):
- `CompanyDomainMatcher` — single source of truth for resolving a domain → `Company` via the team's `domains` custom field; the regex match was previously duplicated across `LinkEmailAction` and `LinkMeetingAction`.
- `AutoCreateCompanyAction` — creates a company **only if the domain isn't already owned**, and serialises resolve-or-create per `(team, domain)` with a Postgres transaction-level advisory lock (`pg_advisory_xact_lock`). First holder creates; the rest re-check inside the lock and reuse it.
- `UsesTransactionalLock` trait — wraps `DB::transaction` + the advisory lock, consistent with the person-creation path.

Advisory lock over `Cache::lock` because the guarded resource is a Postgres row: no TTL to outrun, released on commit/rollback, no second system to fail over. A unique constraint isn't viable since the domain lives in `custom_field_values.json_value` (JSON).

Added regression tests for the duplicate-domain and reuse-existing cases.

AsmitNepali:
Follow-up fix merged into this branch via #388: **auto-created People were keyed on the From display name instead of the email address.**

`AutoCreatePersonAction` used `updateOrCreate` keyed on `name` (+ team/company/source). Since email identity lives in the people `emails` custom field (EAV), not a column, keying on name was wrong both ways:
- **Over-merge** — two distinct people sharing a display name (e.g. two "John Smith" at different domains) collapsed into one record.
- **Under-dedupe** — the same person whose display name varied between emails ("John Smith" vs "John") spawned a second record.

Fix: email-based **find-then-create** (`findByEmail` matches the `emails` custom field → reuse or `create()`), wrapped in a transaction-scoped advisory lock keyed on `team:email` so concurrent sync workers can't double-insert. The Postgres `pg_advisory_xact_lock` primitive is isolated behind an `App\Support\Database\AdvisoryLock` seam. Covered by two new `LinkEmailActionTest` cases (distinct-names → 2 records; same-address → 1 record).

AsmitNepali:
Merged #392 into this branch.

**Adds:** email detail header reply/reply-all/forward controls are now a single native Filament `ActionGroup` dropdown placed inline beside the sent date, triggered by a vertical three-dot (ellipsis) icon — replacing the previous stacked Alpine icon buttons.

- Dropdown items (Reply / Reply All / Forward) are real Filament actions → native modal integration, theming, accessibility.
- Reply/forward logic consolidated into a single builder + helpers in `HasEmailComposeActions`; the legacy email-thread dispatch path reuses them (no duplicated form/send logic).
- Grouped-action triggers can't carry per-action arguments, so the target email resolves from the open email (`selectedEmailId`) — covered by a new `ReplyForwardEmailTest` case.

Header UI + internal refactor only; no change to what reply/forward send. Gates: Pint, PHPStan (0), type-coverage 100%, ReplyForwardEmailTest 6/6.

AsmitNepali:
Merged #391 into this branch — adds **Mark all emails as read** to the inbox.

Resolves the bulk side of the Test Plan's "mark-as-read works" item: a **Mark all read** action in the inbox left panel clears every visible unread inbound email in the current folder for the viewer in one click (per-viewer `email_reads`, single bulk insert, respects `VisibleEmailScope`, no-op on Sent). Covered by 3 new `EmailUnreadCountPerViewerTest` cases (marks-all, skips non-visible, idempotent). pint / phpstan / type-coverage 100% green.

AsmitNepali:
Merged #393 into `feat/email-calendar-integration`. Summary of what it changed:

**Email auto-labeling: LLM → deterministic rules**
- Removed `ClassifyEmailJob` — every inbound email no longer triggers a per-message Anthropic call (it was misusing `services.anthropic.summary_model` to emit a one-word label), eliminating per-email cost, queue latency, and sending the message body to a third party.
- Added `EmailClassifier` — pure heuristics run inline during sync: provider-native category → `.ics`/calendar part → role-addressed senders (`billing@`, `support@`, `no-reply@`) → subject keywords → internal → `Other` fallback. Ambiguous mail lands in `Other` rather than being guessed.
- `EmailCategory` is now a proper Filament enum (`HasLabel`/`HasIcon`/`HasColor`), living in the package next to the other `Email*` enums.

**Inbox UI**
- Category shown as a subtle colored dot + muted label (was a filled pill that dominated each row), using Filament's own color system.

Tests/gates: pint, rector, phpstan, 100% type coverage, email + arch suites pass.

AsmitNepali:
Merged #394 into `feat/email-calendar-integration`.

**Resolves:** duplicated email-sync configuration. Sync settings (`initial_days`, `batch_size`, `interval`) were defined in **two** places — `config/services.php` (`email_sync`) and `packages/EmailIntegration/config/email-integration.php` (`sync`) — both reading the same `EMAIL_SYNC_*` env vars. `InitialEmailSyncJob` read inconsistently across the two files (`initial_days` from one, `batch_size` from the other), and the `interval` keys were dead (the schedule hardcodes `everyFiveMinutes()`).

Now a single source of truth in `email-integration.sync`. Removed the duplicate `services.email_sync` block and the dead `interval` keys. Verified: Pint clean, `SyncQueueRoutingTest` passing.

AsmitNepali:
**#395 merged into this branch** — renders the email detail-bar actions (**Summarize Thread**, **Sharing**, **Request Access**) as native Filament actions instead of hand-rolled `<button>`s calling `$wire.mountAction`.

Resolves: those buttons gave no in-flight feedback. Summarize Thread fires an LLM request that can take seconds with no spinner/disabled state, since Filament's automatic loading indicator only wires up on its own action triggers (the mechanism Reply / Reply All already use). Now each swaps its icon → spinner and disables while the action mounts, scoped per-email via `wire:target`.

Touches Test Plan item: *"sharing + summarize + request-access row actions still work."*

AsmitNepali:
Folded in #396 (merged to `feat/email-calendar-integration`): **per-account inbox filter**.

Before, the inbox merged messages from every connected account with no way to scope to one — so multi-account users couldn't tell which mailbox they were looking at. Now:
- Inbox lands on the user's **default account first** (instead of a merged stream), consistent with how compose/reply already force an account choice.
- An account **switcher** above the folder tabs allows instant jumping between accounts, plus an **"All accounts"** option for the unified view.
- Switcher hidden for single-account users; list + unread badge respect the active scope; `?account=` is validated against the user's own accounts (no cross-user leak).

Covers the multi-account gap in the inbox portion of this PR's test plan.

AsmitNepali:
Merged #399 into `feat/email-calendar-integration` — hardens the email-sync auto-create path on this branch. No-reply / automated senders (`notice@`, `noreply@`, `bounce@`) no longer spawn junk Company/Person records, and auto-created company names now resolve the registrable domain via the Public Suffix List (`email.anthropic.com` → Anthropic, `acme.co.uk` → Acme) instead of leaking mail subdomains/TLDs. Email-integration config also moved to app-level `config/`. Already on this branch — no rebase needed.

AsmitNepali:
Merged #400 into `feat/email-calendar-integration` — brings the Company → Emails / People → Email record inbox to parity with the main Mail inbox: adds the unread-count + "Mark all read" bar and fixes the email-detail loading spinner that rendered top-left instead of centered. Mark-all-read is scoped to the record's own emails; main inbox path unaffected.

AsmitNepali:
Merged #401 into `feat/email-calendar-integration`, which lands on this branch — new connected accounts now default `contact_creation_mode` to **Bidirectional** instead of **All**, so auto-person-creation only fires on two-way correspondence and the CRM no longer fills with one-off contacts from every inbound/outbound address.

AsmitNepali:
Merged **#402** into `feat/email-calendar-integration` — folds into this PR:

- Inbox compose now defaults the **From** account to the inbox-selected account, instead of always the first active account.
- Falls back to the first active account when viewing "all" / no specific account.
- Reply/forward unchanged — still default to the account that received the email.

AsmitNepali:
Merged **#404** into `feat/email-calendar-integration` — folds into this PR:

- Wires the **Sync inbox** / **Sync sent** toggles to the sync path — they persisted but were ignored, so all mail imported regardless of direction.
- `StoreEmailJob` now skips any direction the user toggled off; covers Gmail + Microsoft, initial backfill + incremental.
- No default change: both toggles default `true`. Forward-only — already-synced mail untouched.

AsmitNepali:
## ✅ QA pass — email + calendar integration

Reviewed end-to-end on a fresh checkout of this branch, driving the real app in a browser as a team owner with connected Gmail accounts and seeded mail. **All 310 EmailIntegration tests pass (854 assertions).**

### Browser-verified surfaces (all working)
- **Inbox** — account switcher, All/Inbox/Sent tabs, search, rendered HTML reading pane, categories, unread count, mark-all-read, pagination
- **Accounts** — connected accounts with Active/Default badges, Connect Gmail (Outlook intentionally hidden)
- **Email Templates** — created via the modal → persisted scoped to the current team (team-scoping fix correct end-to-end)
- **Signatures**, **Outbox** (status tabs + columns), **Workspace Privacy** (sharing tier, auto-hide internal, protected recipients)
- **People → Emails** contextual inbox (CRM ↔ email link), **Compose** modal, **Meetings** list
- No JS console errors attributable to this PR.

**Not verifiable in a sandbox (no live OAuth) — covered by passing tests:** real Gmail sync/send/receive, calendar/meeting display, Microsoft Graph redirect (scopes, `prompt=consent`, calendar scope). Outlook connect button deliberately hidden for now (`EmailAccountsPage.php:85`).

---

### 🚀 Production enablement steps

**1. Build & migrate** (new Composer deps land here — without `composer install`, `Socialite::driver('azure')` fails and `MicrosoftOAuthRedirectTest` errors; CI is fine, installs from the lock):
```bash
composer install --no-dev          # adds socialiteproviders/microsoft-azure, manager, google/apiclient
php artisan migrate --force
npm run build
```

**2. Enable the feature flag** (global — turns it on for all teams; resolves from `config('relaticle.features.email_integration')`, *not yet in `.env.example`*):
```dotenv
RELATICLE_FEATURE_EMAIL_INTEGRATION=true
```
```bash
php artisan config:cache
```

**3. Gmail OAuth** (Google Cloud: enable Gmail API + Calendar API, add the redirect URI, request `gmail.readonly` / `gmail.send` / `calendar.readonly` scopes):
```dotenv
GOOGLE_CLIENT_ID=...
GOOGLE_CLIENT_SECRET=...
GOOGLE_REDIRECT_URI=https://<app-domain>/email-accounts/callback/gmail
```
*(Outlook, when un-hidden: `AZURE_CLIENT_ID` / `AZURE_CLIENT_SECRET` / `AZURE_REDIRECT_URI` / `AZURE_TENANT_ID`.)*

**4. Queues — Redis + Horizon** (supervisors for `emails-sync`, `emails-priority`, `emails-bulk` already in `config/horizon.php`):
```dotenv
QUEUE_CONNECTION=redis
```
```bash
php artisan horizon          # or restart it on deploy: php artisan horizon:terminate
```

**5. Scheduler** (drives sync + outbox; all three tasks self-register, no manual schedule edits):
```bash
php artisan schedule:run
```
Registered tasks: `email-incremental-sync` (5 min), `calendar-incremental-sync` (5 min), `email:dispatch-outbox` (every min, honors per-account hourly/daily send limits).

**6. Smoke-test in prod:** connect a Gmail account on the Accounts page → `InitialEmailSyncJob` backfills ~90 days → mail appears in Inbox and on matching People/Company/Opportunity records.

Verdict: **production-ready** 🟢 — code and tests are green; remaining items are infra config (OAuth creds, queue worker, feature flag), not code.

PR #239, opened 2026-04-24
Fix/array validation n plus 1

ManukMinasyan:


ManukMinasyan:
Addressed Copilot's note about query log determinism in 7b3ac431 by flushing after `DB::enableQueryLog()` everywhere we measure query counts. CI re-run incoming.

PR #256, opened 2026-04-29
fix: deployable compose.yml + trust private-network proxies

### LINKED ISSUE #250: Deploying on Coolify
### Relaticle Version

latest

### PHP Version

latest

### Database

PostgreSQL 17+

### Deployment Method

Other

### Problem description

Following https://relaticle.com/docs/self-hosting#deploying-on-coolify leads to:

```
target scheduler: failed to solve: failed to read dockerfile: open Dockerfile: no such file or directory
```

probably bc it's trying to build. If you guys have pre-built images, it needs to be updated in the compose and remove the build section

### Expected behavior

Deploy

### Steps to reproduce

Follow the steps in https://relaticle.com/docs/self-hosting#deploying-on-coolify

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ManukMinasyan:
## Summary

Closes #250.

Two related fixes for self-hosted deploys (Coolify, Dokploy, anyone curl-ing the compose.yml standalone):

- **`compose.yml`** — removed the `build:` blocks from `app`, `horizon`, and `scheduler`. They referenced a Dockerfile that isn't in the build context when the file is pulled by itself, which is exactly the failure reported in the issue (`failed to read dockerfile: open Dockerfile: no such file or directory`). Pre-built multi-arch images already publish to `ghcr.io/relaticle/relaticle` via `.github/workflows/docker-publish.yml`, so the `image:` directive is sufficient.
- **`bootstrap/app.php`** — registered `$middleware->trustProxies(at: [...])` with RFC1918 + loopback + IPv6 ULA/link-local CIDRs. Without this, Laravel's default behavior is to ignore `X-Forwarded-Proto` from any proxy, which causes `request()->isSecure()` to return false behind Coolify's Traefik. The visible symptom is the unstyled login page the reporter shared in the issue comment — Vite emits `http://` asset URLs on an HTTPS page and the browser blocks them as mixed content.

## Why these CIDRs

The list trusts proxies on private networks only:

```
127.0.0.0/8           IPv4 loopback
10.0.0.0/8            RFC1918
172.16.0.0/12         RFC1918 (Docker default bridge range)
192.168.0.0/16        RFC1918
169.254.0.0/16        link-local
::1/128               IPv6 loopback
fc00::/7              IPv6 ULA
fe80::/10             IPv6 link-local
```

Covers Coolify/Dokploy/Traefik on a Docker network, host-local nginx/Caddy, and direct-loopback proxies. Public-internet IPs are rejected — if the FPM port is ever reachable directly (which the standalone `compose.yml` allows via `ports: ${APP_PORT:-80}:8080`), an attacker still can't spoof `X-Forwarded-*` headers.

This is the canonical Laravel 12 idiom — matches the example in https://laravel.com/docs/12.x/requests#configuring-trusted-proxies. No new config file, no env plumbing, no service provider mutation.

## Test plan

- [x] `tests/Feature/Middleware/TrustProxiesTest.php` covers three regression cases:
  - private-network proxy → forwarded headers honored, `isSecure() === true`
  - public-IP caller → forwarded headers rejected (security regression test)
  - loopback proxy → forwarded headers honored
- [x] `vendor/bin/pint --dirty --format agent` — passed
- [x] `vendor/bin/phpstan analyse bootstrap/app.php` — 0 errors
- [x] `vendor/bin/rector --dry-run` — clean
- [x] `composer test:type-coverage` — 100.0%
- [x] `php artisan test --compact tests/Feature/Middleware/ tests/Feature/Auth/` — 31 passed
- [x] `docker compose -f compose.yml config --quiet` — valid
- [ ] Verify on a real Coolify deploy: paste compose.yml, set `APP_KEY` / `DB_PASSWORD` / `APP_URL=https://...`, deploy, confirm login page renders styled

### DIFF
--- a/bootstrap/app.php
+++ b/bootstrap/app.php
@@ -39,6 +39,17 @@ return Application::configure(basePath: dirname(__DIR__))
         },
     )
     ->withMiddleware(function (Middleware $middleware): void {
+        $middleware->trustProxies(at: [
+            '127.0.0.0/8',
+            '10.0.0.0/8',
+            '172.16.0.0/12',
+            '192.168.0.0/16',
+            '169.254.0.0/16',
+            '::1/128',
+            'fc00::/7',
+            'fe80::/10',
+        ]);
+
         $middleware->prepend(SubdomainRootResponse::class);
 
         $middleware->prependToPriorityList(
--- a/compose.yml
+++ b/compose.yml
@@ -1,9 +1,5 @@
 services:
   app:
-    build:
-      context: .
-      dockerfile: Dockerfile
-      target: production
     image: ghcr.io/relaticle/relaticle:latest
     restart: unless-stopped
     ports:
@@ -57,10 +53,6 @@ services:
       start_period: 60s
 
   horizon:
-    build:
-      context: .
-      dockerfile: Dockerfile
-      target: production
     image: ghcr.io/relaticle/relaticle:latest
     restart: unless-stopped
     command: ["php", "/var/www/html/artisan", "horizon"]
@@ -97,10 +89,6 @@ services:
       start_period: 10s
 
   scheduler:
-    build:
-      context: .
-      dockerfile: Dockerfile
-      target: production
     image: ghcr.io/relaticle/relaticle:latest
     restart: unless-stopped
     command: ["php", "/var/www/html/artisan", "schedule:work"]
--- a/packages/Documentation/resources/markdown/self-hosting-guide.md
+++ b/packages/Documentation/resources/markdown/self-hosting-guide.md
@@ -220,7 +220,7 @@ labels:
   - "traefik.http.services.relaticle.loadbalancer.server.port=8080"
 ```
 
-**Note**: When using a reverse proxy, set `APP_URL` to your public HTTPS URL (e.g., `https://crm.example.com`). The `TRUSTED_PROXIES` environment variable defaults to `*` in Docker, so forwarded headers are handled automatically.
+**Note**: When using a reverse proxy, set `APP_URL` to your public HTTPS URL (e.g., `https://crm.example.com`). The app trusts `X-Forwarded-*` headers from RFC1918 private networks, loopback, and IPv6 ULA/link-local — covering Coolify/Dokploy/Traefik on a Docker network and reverse proxies on the host. Headers from public IPs are rejected, preventing spoofing.
 
 ---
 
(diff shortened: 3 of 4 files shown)


ManukMinasyan:
Addressed both review comments in 5b98883c.

**Comment 1 (host-spoof assertion):** strengthened the public-IP rejection test — it now asserts the resolved host stays on the legit `HTTP_HOST` value (`crm.example.com`) and is *not* the spoofed `X-Forwarded-Host` (`evil.example.com`). Locks in host-header-injection prevention as a regression test alongside the scheme/IP rejection.

**Comment 2 (configurable / narrower trust list):** keeping the hardcoded RFC1918 + loopback + IPv6 private list. Rationale:

- This matches Laravel 12's docs example verbatim — `$middleware->trustProxies(at: ['192.168.1.1', '10.0.0.0/8'])` is the canonical idiom for v12.
- A narrower default (e.g. only `172.16/12`) breaks deploys on Docker Swarm (`10/8`), custom user-defined networks (`192.168/16`), or hosts with `bip` overrides — exactly the "broken CSS behind Traefik" failure mode this PR exists to prevent.
- An env-driven override was tried in earlier iterations and ran into real Laravel constraints: `Env::get()` in `bootstrap/app.php` can't read `.env` after `config:cache` (the `LoadEnvironmentVariables` bootstrapper is skipped); a `config/trustedproxy.php` file would add legacy-fideloper-pattern surface area for one Laravel built-in. Comparable mature self-hosted Laravel apps (Firefly III, BookStack, Cachet, Pterodactyl, Akaunting, Monica) either default to broken-until-configured (causing this exact bug on first deploy) or default to `**` (insecure if FPM port is exposed). The hardcoded private-CIDR default is strictly safer than the surveyed alternatives.
- Realistic threat residue: a peer container on the same Docker network can spoof `X-Forwarded-*`. Mitigations: (a) for queued-job/email URL generation Laravel uses `APP_URL`, not `request()->getHost()`, so the host-injection blast radius is limited to in-request URL generation; (b) Coolify's default deployment routes via Traefik labels and doesn't require the `${APP_PORT}:8080` host port mapping, so operators concerned about untrusted Docker peers can drop that port from their override.

Bonus: added two extra tests (5b98883c) that directly assert `asset()` URL scheme follows the forwarded-proto header — proving end-to-end that this PR is what fixes the unstyled-CSS bug reported in #250.

PR #275, opened 2026-05-07
fix: harden favicon fetch job against worker hangs, malformed URLs, and SSRF

ManukMinasyan:
## Summary

Eliminates the three production errors emitted by `App\Jobs\FetchFaviconForCompany` (`MaxAttemptsExceededException`, `UnreachableUrl` with HTML mixed into the URL, `DiskCannotBeAccessed`) and the redundant-dispatch pattern that amplified them.

Five focused commits, each independently shippable.

### Root causes addressed

1. **`MaxAttemptsExceededException` (5 events / 17h, unhandled).** `favicon-fetcher` config had `timeout: 0` / `connect_timeout: 0`, so a slow remote site hung the worker until Horizon's 60s timeout sigkilled it; the job was requeued, then `attempts() (2) > tries (1)` made the next pickup throw before `handle()` ran. Fix: enforce 10s/5s HTTP timeouts and add a 30s job-level `timeout` so the job throws cleanly into its catch.
2. **`UnreachableUrl` with HTML in the URL (2 events).** `HighQualityDriver::tryHighResFavicon` used `(.*?)` non-greedy capture inside patterns like `/href=["\'](.*?)["\']\s+[^>]*sizes=["\']512x512["\']/`; on minified HTML the engine extended `.*?` across multiple `<link>` tags to satisfy the trailing `sizes="..."`, capturing nested tags as part of the URL. `tryAppleTouchIcon` had the same bug class (no `\b` boundary on `href=`, no symmetric pattern for `<link href="..." rel="apple-touch-icon">`). Fix: anchor patterns with `<link\b[^>]*` and capture URLs with `[^"\']+` so matches cannot escape a single tag.
3. **CPU saturation amplifier.** `CompanyObserver::saved` dispatched `FetchFaviconForCompany` on every Company save, regardless of whether the domain changed; combined with hung HTTP fetches, imports/bulk edits flooded the queue. Fix: short-circuit when `hasMedia('logo')` is true (callers can clear the logo collection to force a refetch).
4. **SSRF surface.** A user-controlled domain field flowed straight into server-side fetches against attacker-controlled URLs. Fix: introduce `SsrfGuard::assertPublicHost()` resolving via `dns_get_record` and rejecting RFC1918/loopback/link-local (incl. IPv6) via `FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE`; wired into 4 sites (`fetch`, `tryGoogleHighRes`, `tryDuckDuckGo`, and the job's `addMediaFromUrl` to close the redirect-to-private-host gap when an attacker-controlled HTML returns `<link href="//127.0.0.1/" rel="apple-touch-icon">`). All catches `report()` so SSRF attempts surface in Flare.

### Defense in depth

- `catch (Exception)` → `catch (Throwable)` in the job so PHP `Error` types (TypeError, etc.) are reported rather than silently escaping.
- `filter_var($url, FILTER_VALIDATE_URL)` belt-and-braces guard before `addMediaFromUrl`.
- Bounded `uniqueFor = 600` on the `ShouldBeUnique` lock so a stuck unique-lock can't block re-dispatch indefinitely.

## Test plan

- [x] 20 new feature tests cover regex anchoring (incl. decoy tags + size preference + apple-touch-icon adversarial input), job property pinning, SSRF rejection at SsrfGuard / driver / job layers, observer skip-when-logo-exists, observer skip-when-no-domain, and `Throwable` swallowing via `Favicon::shouldReceive('driver')->andThrow(new TypeError(...))`.
- [x] `vendor/bin/pint --dirty --format agent` passes.
- [x] `vendor/bin/rector --dry-run` clean (added `UniqueForPropertyToUniqueForAttributeRector` to existing skip list, consistent with prior `Tries`/`Timeout` skips).
- [x] `vendor/bin/phpstan analyse` 0 errors.
- [x] `composer test:type-coverage` 100%.
- [x] All 20 favicon tests pass: `php artisan test --compact tests/Feature/Services/Favicon/ tests/Feature/Jobs/FetchFaviconForCompanyTest.php tests/Feature/Observers/CompanyObserverFaviconTest.php`.

## Notes

- The `hasMedia('logo')` short-circuit is a behavior change: domain edits after a logo has been stored will not re-fetch. There is no UI path for refetch today; documented inline at the early return.
- DNS rebinding window between `SsrfGuard::assertPublicHost()` and the actual HTTP request is open by design — for a queue worker hitting each domain once, attacker payoff is low.
- `DiskCannotBeAccessed` (1 event in 17h) was not addressed in code: storage is correctly mounted in compose.yml, the error reads as a transient IO/permissions blip on the prod volume. Worth monitoring; if it recurs, investigate volume UID/GID alignment between `app` and `horizon` containers.

### DIFF
--- a/app/Exceptions/SsrfGuardException.php
+++ b/app/Exceptions/SsrfGuardException.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace App\Services\Favicon;
+namespace App\Exceptions;
 
 use RuntimeException;
 
--- a/app/Jobs/FetchFaviconForCompany.php
+++ b/app/Jobs/FetchFaviconForCompany.php
@@ -7,12 +7,12 @@ namespace App\Jobs;
 use App\Enums\CustomFields\CompanyField;
 use App\Models\Company;
 use AshAllenDesign\FaviconFetcher\Facades\Favicon;
-use Exception;
 use Illuminate\Contracts\Broadcasting\ShouldBeUnique;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Foundation\Bus\Dispatchable;
 use Illuminate\Foundation\Queue\Queueable;
 use Illuminate\Support\Str;
+use Throwable;
 
 final class FetchFaviconForCompany implements ShouldBeUnique, ShouldQueue
 {
@@ -20,6 +20,12 @@ final class FetchFaviconForCompany implements ShouldBeUnique, ShouldQueue
     use Queueable;
 
     public bool $deleteWhenMissingModels = true;
+
+    public int $tries = 1;
+
+    public int $timeout = 30;
+
+    public int $uniqueFor = 600;
 
     public function __construct(public readonly Company $company) {}
 
@@ -47,6 +53,10 @@ final class FetchFaviconForCompany implements ShouldBeUnique, ShouldQueue
                 return;
             }
 
+            if (filter_var($url, FILTER_VALIDATE_URL) === false) {
+                return;
+            }
+
             $path = parse_url($url, PHP_URL_PATH);
             $extension = $path ? pathinfo($path, PATHINFO_EXTENSION) : '';
 
@@ -70,7 +80,7 @@ final class FetchFaviconForCompany implements ShouldBeUnique, ShouldQueue
                 ->toMediaCollection('logo');
 
             $this->company->clearMediaCollectionExcept('logo', $logo);
-        } catch (Exception $exception) {
+        } catch (Throwable $exception) {
             report($exception);
         }
     }
--- a/app/Jobs/FetchFaviconForCompany.php
+++ b/app/Jobs/FetchFaviconForCompany.php
@@ -6,6 +6,8 @@ namespace App\Jobs;
 
 use App\Enums\CustomFields\CompanyField;
 use App\Models\Company;
+use App\Services\Favicon\SsrfGuard;
+use App\Services\Favicon\SsrfGuardException;
 use AshAllenDesign\FaviconFetcher\Facades\Favicon;
 use Illuminate\Contracts\Broadcasting\ShouldBeUnique;
 use Illuminate\Contracts\Queue\ShouldQueue;
@@ -54,6 +56,14 @@ final class FetchFaviconForCompany implements ShouldBeUnique, ShouldQueue
             }
 
             if (filter_var($url, FILTER_VALIDATE_URL) === false) {
+                return;
+            }
+
+            try {
+                SsrfGuard::assertPublicHost($url);
+            } catch (SsrfGuardException $exception) {
+                report($exception);
+
                 return;
             }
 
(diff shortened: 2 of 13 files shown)


ManukMinasyan:
Thread #1 (`ShouldBeUnique` interface): addressed in 491368af. Confirmed by reading vendor source — `Illuminate\Queue\Queue::push` and `CallQueuedHandler::commandShouldBeUnique` both check `instanceof Illuminate\Contracts\Queue\ShouldBeUnique`, so the broadcasting variant we previously imported was a no-op. `uniqueId()` and `uniqueFor` now actually fire.

Thread #2 (`SerializesModels`): not a real bug — `Illuminate\Foundation\Queue\Queueable` (already used) is the Laravel 11+ unified trait that bundles `Dispatchable + Bus\Queueable + InteractsWithQueue + SerializesModels` in one. Verified by reading `vendor/laravel/framework/src/Illuminate/Foundation/Queue/Queueable.php`. Closing this thread without code change.

ManukMinasyan:
Thread #6 (`Str::start` double-prepending scheme): addressed in 6c26ad5c. Replaced `Str::start($domainName, 'https://')` with a `Str::startsWith` check so values already containing `http://` or `https://` are preserved. Verified: `http://example.com` → `http://example.com`, `example.com` → `https://example.com`, `https://example.com` → `https://example.com`.

ManukMinasyan:
Thread #3 (SSRF gap in `faviconIsAccessible`): addressed in 17fb260d. Added `SsrfGuard::assertPublicHost($faviconUrl)` before the `HEAD` request inside `faviconIsAccessible()`. The candidate URL there comes from `convertToAbsoluteUrl()` on attacker-controlled HTML (e.g. `href="//127.0.0.1/icon.png"` or `href="http://169.254.169.254/..."`) so it now gets the same private/loopback rejection as the top-level guard.

Note: the apple-touch fallthrough at line 111 (`stripPathFromUrl($url).'/apple-touch-icon.png'`) uses the source URL's host, which was already validated by the top-level `fetch()` guard — no extra check needed there. Verified all 8 driver tests still pass.

ManukMinasyan:
Thread #4 (URL credential leakage in exception messages): addressed in a1604fe6. The first `throw_if` no longer interpolates the full URL — when host parsing fails the URL is malformed by definition and not useful for debugging anyway. The other two messages already only include the resolved `$host` and `$address` (post-DNS resolution, sans userinfo/path/query) so they don't leak credentials. Downstream Flare reports will no longer carry `user:pass@...` if a userinfo URL ever reaches the guard.

ManukMinasyan:
Thread #5 (test flakiness from live DNS): addressed in 41431236. Switched all `https://example.com` URLs in HighQualityDriverTest to `https://1.1.1.1` (Cloudflare's public anycast literal). `SsrfGuard::resolveAddresses()` short-circuits IP literals via `filter_var($host, FILTER_VALIDATE_IP)` (line 33 of SsrfGuard.php), so `dns_get_record` is never called. Tests now run without any DNS dependency. `Http::fake` URLs updated to match. 8/8 driver tests still pass.

PR #321, opened 2026-06-09
fix(chat): harden AI approval flow against lost state and rate limits

ManukMinasyan:
## What & why

The in-app CRM chat assistant broke during a bulk "create 50 tasks one-by-one" flow (from a reported transcript): after approving a proposal it kept insisting *"I still need your approval on Task 1/50"*, then surfaced a generic *"The assistant encountered an error."* Root-caused to two defects plus a silent stall. This fixes all three.

### Fix 1 — Resolved-action context (Bug A: lost approval state)

Approvals were journaled into the LLM transcript **only** by the `ContinueChatMessage` continuation, which `RememberConversation` persists only on a *successful* stream. When the continuation threw (429) or was skipped by the chain cap, the approval never entered the transcript, so the model replayed a history ending at the *pending* proposal and re-asked for approval.

Now a `<resolved_actions>` block is injected into the system prompt each turn, built from the authoritative `pending_actions` table and scoped to "resolved after the last assistant message" (self-limiting, **no schema change**). Mirrors the existing `<superseded_proposals>` plumbing.

### Fix 2 — 429 resilience (Bug B: opaque provider errors)

`RateLimitedException` (429) thrown from `$agent->stream()` propagated to `failed()` and broadcast the generic error with no retry (429 doesn't fail over to the second provider). Both jobs now catch `RateLimitedException`/`ProviderOverloadedException`, `release()` with capped exponential backoff (attempts-gated, bounded by `retryUntil`), and broadcast an honest "being rate-limited" message. `ContinueChatMessage` reserves credit only on attempt 1 to avoid double charges across retries.

### Fix 3 — Explicit pause

When the auto-continue chain hits `CHAIN_HARD_CAP` it previously stopped silently. It now broadcasts a `ChatPaused` event and the UI renders a *"Paused — say continue to resume"* notice; Fix 1 lets the model resume from where it left off.

## Tests

- New: `ResolvedActionsContextTest` (incl. a Bug A regression that fails on `main`), `CrmAssistantResolvedBlockTest`, `ChatRateLimitTest`, `ChainCapPauseTest`.
- Full Chat suite: **451 passed / 1103 assertions**. PHPStan (`packages/Chat`) clean, Pint clean, Rector clean.

## Manual verification still needed

The blade pause-notice handler (`handleChatPaused`) has no automated seam — needs a browser check: approve 5 proposals in a row → confirm the pause bubble appears → "continue" resumes.

## Notes

- Adjusted two test assertions (test layer, not production): the `CrmAssistant` "omits block" check (the new prose legitimately names the `<resolved_actions>` tag) and `TerminalBroadcastInvariantTest`'s brittle fixed-width window.

### DIFF
--- a/packages/Chat/resources/views/livewire/chat/chat-interface.blade.php
+++ b/packages/Chat/resources/views/livewire/chat/chat-interface.blade.php
@@ -898,7 +898,8 @@ Alpine.data('chatInterface', (initialConversationId, sendUrl, initialMessage, in
             .listen('.stream.failed', (e) => this.handleStreamFailed(e))
             .listen('.conversation.resolved', (e) => this.handleConversationResolved(e))
             .listen('.follow_ups', (e) => this.handleFollowUps(e))
-            .listen('.pending_actions_superseded', (e) => this.handlePendingActionsSuperseded(e));
+            .listen('.pending_actions_superseded', (e) => this.handlePendingActionsSuperseded(e))
+            .listen('.chat.paused', (e) => this.handleChatPaused(e));
 
         return readyPromise;
     },
@@ -1431,6 +1432,19 @@ Alpine.data('chatInterface', (initialConversationId, sendUrl, initialMessage, in
         this.queuedSend = null;
         this.clearStreamTimeout();
         this.restoreInputFocus();
+    },
+
+    handleChatPaused(event) {
+        const nowIso = new Date().toISOString();
+        this.messages.push({
+            role: 'assistant',
+            content: event?.message || 'Paused. Say "continue" to keep going.',
+            pending_actions: [], paywall: null, sessionExpired: false,
+            rendered: true, prerendered: false, copiedAt: 0, follow_ups: [],
+            created_at: nowIso,
+        });
+        this.isStreaming = false;
+        this.$nextTick(() => this.restoreInputFocus());
     },
 
     restoreInputFocus() {
--- a/packages/Chat/src/Agents/CrmAssistant.php
+++ b/packages/Chat/src/Agents/CrmAssistant.php
@@ -79,6 +79,15 @@ final class CrmAssistant implements Agent, Conversational, HasMiddleware, HasPro
      */
     public array $supersededProposals = [];
 
+    /**
+     * Actions already resolved (approved/rejected/expired/superseded) since the
+     * last assistant turn, injected so the model knows their outcome even if the
+     * approval continuation never journaled them into the transcript.
+     *
+     * @var list<array{operation: string, entity_type: string, status: string, label: string|null, record_id: string|null}>
+     */
+    public array $resolvedActions = [];
+
     public function withConversationId(?string $conversationId): self
     {
         $this->conversationId = $conversationId;
@@ -112,6 +121,7 @@ For any create, update, or delete operation:
 - The tool returns a pending_action proposal -- do NOT tell the user the action was completed
 - Tell the user you've proposed the action and ask them to review the proposal card above
 - Wait for the user to approve or reject before proceeding
+- If a multi-step sequence pauses, tell the user it paused and that they can say "continue" to resume; then resume from the resolved actions when they do
 
 ## Formatting
 - Use markdown for rich text formatting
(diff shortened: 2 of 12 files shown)


ManukMinasyan:
## 🧪 Business review — PR #321 (real-browser E2E)

**Verdict: ✅ approved with 2 pre-deploy conditions · 0 defects found**
Driven through the live app (`app.relaticle.test`) with **real OpenAI + Reverb streaming**. (Note: an initial "degraded channel" read was a stale cached JS bundle + a wrong-host websocket from another Herd site — cleared by a rebuild; streaming then worked.)

### Verified end-to-end ✅
| AC | Result | Evidence |
|---|---|---|
| **Regression** — chat works, AI streams live | ✅ PASS | "CRM overview" reply rendered live in the browser |
| **Fix 1 (Bug A)** — create → approve → record created, **no stuck re-approval** | ✅ PASS | `br-321-fix1check` + `br-321-p1..p5` created (DB-confirmed); success messages streamed; assistant never re-asked for an approved action |
| **Fix 3 — pause notice renders** | ✅ PASS | a real `ChatPaused` broadcast to the conversation channel rendered *"Paused… say continue"* in the browser within 2s (this is the handler that had **no automated test**) |
| New Fix-3 code ships in served page | ✅ PASS | `.chat.paused` listener + `handleChatPaused()` present |

### Notable corroboration of Fix 1 👍
During a rapid 6-task one-at-a-time run, the **Bug A desync actually occurred** — 6 approvals landed in `pending_actions` while the transcript journaled fewer (continuation journaling lagged) — **yet the UI stayed correct throughout**. That's precisely the desync the new `<resolved_actions>` block is designed to tolerate. Good real-world signal that Fix 1 does its job.

### Could not be verified locally (env limits, not PR defects) ⚠️
1. **Fix 2 (429 retry/backoff + no double charge)** — can't force a real provider 429, and local `QUEUE_CONNECTION=sync` doesn't exercise `release()`/retry the way prod (async queue) will. Covered by `ChatRateLimitTest`.
2. **Fix 3 full 5-deep auto-chain → cap** — under sync queue the auto-continuation chain stalls (WithoutOverlapping lock contention on rapid approvals), so the cap never trips organically. Cap logic covered by `ChainCapPauseTest`; UI handler verified via the direct broadcast above.

### 🚦 Pre-deploy conditions (for 100k prod)
- [ ] Verify **Fix 2** retry-with-backoff + single-charge on **staging with the real async queue driver** (sync hides this path).
- [ ] Verify **Fix 3** full auto-chain pause (≥5 sequential auto-approvals → notice) on staging/async queue.
- [ ] (Low) Glance at per-turn token cost from `<resolved_actions>` at scale — bounded by `limit(20)` + the since-last-assistant window, so low risk.

**Test data left in DB** under the `br-321-` prefix (6 tasks + chat conversations).

<sub>br-sha:091bf0fb</sub>

ManukMinasyan:
## 🔴→✅ Follow-up: real 429 resilience gap found under async queue, now fixed (`f479d40b`)

Per the earlier review's pre-deploy condition, I set up a **real async queue (Redis) + Horizon** locally and stress-tested the chat under load. This surfaced a gap that sync-queue and the unit tests could not.

### The gap
Fix 2 caught `RateLimitedException | ProviderOverloadedException`. But under real load the provider's **streaming** path throws a raw `Illuminate\Http\Client\RequestException` (status 429) — *not* the typed exception (the failover wrapper only normalizes the non-streaming path). So real streaming 429s **bypassed the retry**, failed the job at `attempts=1`, and showed the **generic** error — i.e. the original Bug B symptom still occurred.

**Evidence:** Horizon `failed_jobs` on the `chat` queue during a bulk-approval burst: `{"Illuminate\\Http\\Client\\RequestException":6,"Error":1}`, all status 429.

### The fix (`f479d40b`)
Shared `isRateLimited()` helper now also matches a `RequestException` with status `429/529/503` (status-gated so genuine 4xx errors still fail fast). Both the retry catch and `failed()`'s message branch use it. Added unit coverage asserting `isRateLimited(RequestException(429))` and the streaming-429 failure message.

### Verified E2E (Horizon)
- **Before:** identical burst → 7 jobs failed immediately (`attempts=1`).
- **After:** the same burst → chat jobs now show **`attempts=5`** (released + retried with backoff) before any failure; several completed. The retry path engages on the real streaming 429.
- They only *exhaust-then-fail* here because my test API key is rate-limit-exhausted from a full session — a transient 429 in prod succeeds on retry. On final exhaustion the user now gets the honest "being rate-limited" message, not the generic one.
- Full Chat suite: **453 passed**; PHPStan/Pint/Rector clean.

### Status of the two pre-deploy conditions
- ✅ **Fix 2 (429 retry)** — now genuinely works for the real streaming path; verified under async/Horizon.
- ⚠️ **Fix 3 full 5-deep auto-chain** — still not reproduced organically (the auto-chain tangles under repeated rapid approvals — duplicate proposals + lagged journaling). The UI handler is verified (direct broadcast) and the cap logic is unit-tested, but the *organic* end-to-end pause is worth one more look on staging.

<sub>br-sha:f479d40b</sub>

PR #368, opened 2026-06-20
Merge same-save activity rows into one timeline entry

AsmitNepali:
## Problem

A single update fragments the activity timeline: one native `updated` row plus a separate row **per** custom field, each shown as its own entry. A one-time "Update" looked like five edits (see the screenshot in the issue — `LinkedIn`, `Domains`, `ICP`, native `Name`, all as separate cards).

## Fix (read-side merge keyed on spatie `batch_uuid`)

- **Install activity-log [PR #32](https://github.com/relaticle/activity-log/pull/32)** — adds opt-in `fromActivityLog(mergedRenderer: …)` that collapses rows sharing a non-empty `batch_uuid` into one `TimelineEntry`, unioning their payloads. Pinned to the dev branch (`dev-activity-timeline-batch-merge as 1.2.0`) via a VCS repo until the PR releases.
- **Stamp one `batch_uuid` per request/job** — `Activity::beforeLogging` hook + a request-scoped `RequestActivityBatch` holder, so every row written while handling one request (native column changes *and* the custom-field observer's rows) shares the same id. No change to how/when activities are logged.
- **Host-owned `batch_uuid` column** — guarded `add_batch_uuid_to_activity_log_table` migration (the package only reads it).
- **Unified renderer + blade** — `MergedActivityRenderer` + `merged-activity.blade.php` render native columns and custom fields in one expandable diff card. Wired on all 5 timeline models (`Company`, `Note`, `Opportunity`, `People`, `Task`).
- **Removed** the now-superseded `CustomFieldChangesRenderer` and its blade.

## Verification

- Browser-confirmed in the running app: a combined Name + ICP edit now renders a single **"changed 2 attributes"** entry that expands to `NAME: Airbnb → Airbnb (verified)` and `ICP: Yes → No`.
- `pint` / `rector` clean · `phpstan` 0 errors · type-coverage 100% · Arch + ActivityLog suites green (`BatchMergedTimelineTest` covers merge, no-merge across batches, and null-batch legacy rows).

image

## Notes / caveats

- Merge is per **request**, not per save — two distinct edits to the same record in one request would collapse together (cross-record is unaffected; different subjects have separate timelines).
- Legacy rows (`batch_uuid = null`, written before this change) never merge — only new saves collapse.
- `fromActivityLogOf()` (related-model activity) is **not** merged yet — tracked as a follow-up in PR #32.
- Repin the composer dependency to a tagged `^1.2` once activity-log PR #32 is released; do not ship the dev branch to production.


ManukMinasyan:
@AsmitNepali — pushed a follow-up fix on top of this PR (`042278be1`).

**Issue found in review:** the merged renderer governs *every* event type, so single-row **created** / **deleted** entries rendered as "*X changed N attributes*" with a diff table that exposed internal system columns (`email_count`, `last_email_at`, `avg_response_time_hours`, …) the user never edits. On create that's the first and most prominent timeline entry — on every Company / Person / Opportunity / Task / Note.

**Fix:** the entry label is now operation-aware — only updates (and custom-field-only saves) read as "changed" with a diff; created/deleted/restored keep their own verb and no longer dump the attribute set. Added render-level regression tests for both paths.

Verified in the browser: a just-created record now reads "*X created \<name\>*"; updates still merge native + custom fields into one "changed N attributes" entry. Could you pull and test on your side — create a record, then edit it touching a native + a custom field, and confirm the timeline reads right (created vs changed)?

PR #699, opened 2026-09-10
feat(files): private media uploads and MCP agent uploads

### LINKED ISSUE #686: Route every durable user file through medialibrary collections
## Why

Five upload paths exist today and only one of them goes through medialibrary. Each of the other durable paths stores a bare path string with no owner, no size or type record, and no way to move to a private disk without touching every consumer.

| Upload path | Storage today | Medialibrary |
|---|---|---|
| Company logos | `logo` collection, filled by `FetchFaviconForCompany` | yes |
| Custom field `file-upload` values | Filament `FileUpload` in the custom-fields package, public disk, `uploads/custom-fields`, path string in the value | no |
| Inline attachments in rich editors (note body, task description) | Filament RichEditor `attachFiles`, public disk, `attachments` | no |
| User profile photos | Jetstream, public disk, `profile-photos`, path on `users` | no |
| Import wizard CSVs | `storage/app/imports/{id}`, transient | no, correctly |

## Rule

Durable user files go through medialibrary with a named collection on the owning model. Transient files (import CSVs) and framework-owned files (Jetstream profile photos) are exempt. The rule lands in `.ai/rules/file-uploads.md` with the agent upload work, which makes agent uploads the second consumer (`Team`, `agent-uploads` collection).

## Work

1. Custom field `file-upload`: replace the package `FileUpload` component with a media-backed component so each value has a Media row on the record's model. Lives in the custom-fields package; the value stays a path string for compatibility until every consumer reads the Media row.
2. Rich editor inline attachments: register a medialibrary-backed file attachment provider so pasted images get a Media row on the record being edited.
3. Once both land: flip the `public` disk to a private disk plus signed URLs behind one config change, and serve non-image types with `Content-Disposition: attachment`.

## Done when

- Every durable upload path in the table reads "yes".
- One config change moves all of them to a private disk.
- `tests/Arch` fails on a new `FileUpload::make` or `attachFiles` use outside medialibrary.

ManukMinasyan:
## Summary

Every durable user file is a medialibrary row on the model that owns it, stored on a private disk by default and served through a signed route. On that storage model, rich editor images become media on the record and AI agents get an upload contract over MCP.

Closes #686. Plan B of the agent custom-field work; Plan A (field values over API, MCP and chat) merged in #698.

Design: `docs/superpowers/specs/2026-09-12-durable-files-and-agent-uploads-design.md`. Standing rules: `.ai/rules/file-uploads.md`.

### Storage

- `media` gains one ulid column, `workspace_id` (indexed). Tenant scoping never reads a JSON property.
- A stored reference to a file is the media `uuid`. Paths and URLs derive from the row.
- Every upload lands in the workspace's `pending-uploads` collection. Saving a record whose value references it moves the row in place onto the record's `attachments` collection. Attachments no rich-editor value on the record still names are released after commit. `Media::move()` is never used; ownership changes are attribute writes, so `uuid` and path never change.
- An attachment belongs to the **record**, not to one field. The same image may appear in two rich-editor fields and survives until no value names it. A release reads every rich-editor value on the record and is scoped to the `attachments` collection, so a `logo` row can never be swept.
- `Company`, `People`, `Opportunity`, `Task` and `Note` now implement `HasMedia`. All five soft-delete, so medialibrary releases the files on a force delete; a soft-deleted record keeps them.
- `UploadPathGenerator` keeps the default layout for `logo` (the 9,264 existing company logos stay put) and uses `uploads/{uuid}/` for everything else.
- `MEDIA_DISK` names a disk from `config/filesystems.php` and defaults to `local`. `logo` collections pin `public` on the collection itself.
- `media` joins `Workspace::RESERVED_SLUGS`, so no workspace slug can shadow the serving route.

### There is no `file-upload` custom field type

The type is listed in `->disabled()` in `config/custom-fields.php` and its components are removed. Rich-editor attachments and the MCP upload tools are the only ways a user file reaches a record. `CustomFieldType::FILE_UPLOAD` stays as a case so stored legacy rows still resolve and code that excludes the type compiles.

### Panel

- Rich editor attachments go through `RichContentAttachments`, a Filament `FileAttachmentProvider`. `data-id` is the media uuid. The record page renders through `RichContentEntry`, and REST and MCP reads rewrite `src` from the row.
- A rendered link is signed and expiring; the **stored** body keeps the unsigned `/media/{uuid}`, so a body does not rot when a signature expires. This holds on every write path, including a panel save of a body that was rendered with a signed URL.
- `CustomFieldInput::richText()` tags an untagged `` whose URL names an upload the workspace owns, so markdown an agent wrote claims its file on save.
- `media:backfill-rich-editor-attachments --force` migrates the 15 legacy public-disk images onto their records. Dry run by default, idempotent, and it warns and continues on a row it cannot migrate rather than aborting the run.

### Serving

- `GET /media/{uuid}` is a 30-minute signed route, 300 requests per minute. Images stream inline, everything else downloads under its original name, with `Cache-Control: private, no-store`, `X-Content-Type-Options: nosniff` and a CSP sandbox. A missing file on the disk answers 404, an unsigned or tampered request 403. A name with no ASCII characters still downloads, through the RFC 5987 `filename*` parameter.
- The route is exempt from `RedirectToPrimaryHost`, because a cross-host 301 would invalidate the signature.
- A row on a public disk keeps its plain `/storage` URL, rewritten to the requesting host for same-origin previews. `MediaUrlGenerator` replaces `SameOriginUrlGenerator`.

### Validation

`UploadClaims::assertClaimable()` runs from the `saving` observer hook: a referenced uuid must be pending in the caller's workspace, or already an attachment on this record. It takes a `uuid`-ordered `lockForUpdate`, so a race between two saves cannot persist a value naming a file another record claimed.

### Agent uploads over MCP

- `upload-file` takes exactly one of `source_url`, `base64` plus `filename`, or `upload_id`, and returns `{file_id, name, url, mime_type, size, suggested_markdown}`. `url` is signed and expires. `suggested_markdown` carries the stable, unsigned `/media/{uuid}` link that the write path retags on save.
- `create-upload-url` returns a 5-minute signed `PUT` URL for bodies the agent holds locally. The receiver is signed and throttled at 60 a minute, refuses a missing or oversized `Content-Length` before reading, and deletes the file if the written body still exceeds the ceiling.
- `StoreAgentUpload` fetches URLs through `SsrfGuard::pinnedClient()`: https on 443 only, host resolved once, private ranges refused, connection pinned to the checked address, no redirects, decompression off, 10 MB ceiling. All sources are sniffed with `finfo` against the allowlist: pdf, doc, docx, xlsx, pptx, jpeg, png, gif, webp.
- 60 `upload-file` calls per hour per workspace through `RateLimiter::increment`, refused files included.
- `app:purge-pending-uploads` runs hourly and deletes pending rows and signed-PUT temp files older than 24 hours.

### Deployment

The legacy-image backfill is queued from a migration with `Artisan::queue(...)->onQueue('imports')->afterCommit()`, which is the only shape that reaches a self-hosted install, where an upgrade is `docker compose pull && up -d` and nothing but migrations runs. The command never throws out of `handle()`, because a throw under `QUEUE_CONNECTION=sync` would escape `runMigration()` before the migrator logs the row and re-run on every container start.

### Housekeeping and published copy

- `SsrfGuard` and `HostResolver` moved to `App\Support\Http`, shared by favicon fetching and uploads.
- `tests/Arch/ConventionsTest.php` fails on a new `FileUpload::make(` or attachment disk configuration outside the rich editor field type and the profile photo form.
- Tool count goes from 37 to 39 everywhere it is published: the MCP guide (which also gains a Files section), billing copy, `CompetitorFacts`, the home FAQ, the meta and og descriptions, the schema.org `featureList`, the hero, both pricing views, and the EspoCRM comparison opening.

## Tests

| Area | File |
|---|---|
| Rich editor attachments and read rewrite | `tests/Feature/Filament/App/Resources/RichEditorAttachmentTest.php` |
| Both MCP tools, every source, rate limit, SSRF refusals | `tests/Feature/Mcp/UploadToolsTest.php` |
| Paths, claim and release, signed serving, purge | `tests/Feature/Media/{MediaStorageTest,UploadClaimsTest,PrivateDiskTest,PurgePendingUploadsCommandTest}.php` |
| Legacy image backfill: dry run, force, idempotency, resilience | `tests/Feature/Media/BackfillRichEditorAttachmentsCommandTest.php` |
| Write and include surfaces agree | `tests/Feature/CRM/SurfaceParityTest.php` |
| SSRF ranges `filter_var` calls public | `tests/Feature/Support/Http/SsrfGuardTest.php` |

## Known follow-ups

- `create-upload-url` has no per-workspace budget of its own; the 60/hour limiter is on `upload-file`, and `tmp/` is swept after 24 hours.
- The receiver's pre-read `Content-Length` guard is not isolated by a test: the post-write check returns the same 413.
- Deployment check: a 10 MB base64 payload is a ~14 MB request body, so the MCP host needs `client_max_body_size` and `post_max_size` at or above that. The shipped docker image allows 100M.


ManukMinasyan:
## Pre-merge review: durable files and agent uploads

Verdict: the design holds up and the agent path works end to end. I found and fixed four defects, one of which would have written expiring URLs into customer note bodies on every save. All gates are green. Two items need your call before merge.

All six commits are on this branch, up to `f0f33a89d`.

### Defects found and fixed

**1. The panel persisted an expiring signed URL into every rich-editor body.**
`RichEditorStateCast::get()` only strips an image `src` when `getFileAttachmentsVisibility() === 'private'`. Our custom-field editor never set it, and the default is `'public'`, so the signed preview URL dehydrated straight into `text_value`. Opening a note and pressing Save with no edits rewrote the body with a fresh signature, which also logged a `custom_field_changes` activity entry the user never caused.

Reproduced first, then fixed. Stored shape before and after:

```
- <p><img src="…/media/{uuid}?expires=1789491254&amp;signature=63b6e28…" data-id="{uuid}"></p>
+ <p><img data-id="{uuid}"></p>
```

**2. Document links had the same bug, and the Filament flag does not cover them.**
Neither the state cast nor `RichContentRenderer` touches `a[href]`, so an agent-created `[brief.pdf](/media/{uuid})` link got a signed href persisted on the first panel save and re-signed on every save after. The REST API round-trip had it too: `GET` returns signed URLs, and `PATCH`ing that body back unchanged stored the signature.

Fixed by replacing `tagOwnedAttachments()` with `canonicalize()`, the explicit inverse of `rewriteAttachmentUrls()`. It tags owned images, drops a resolvable `src`, and rewrites an owned href back to the unsigned route. Both write paths call it: `CustomFieldInput::richText()` for API, MCP and chat, and `->dehydrateStateUsing()` for the panel.

Two mechanisms now guard this, and they cover different sets, which I verified by removing each in turn:

- `canonicalize()` does the real work on every write path. With the Filament flag removed, all 15 other tests still pass.
- `->fileAttachmentsVisibility('private')` covers the one case `canonicalize` deliberately leaves alone: an `img[data-id]` this workspace cannot resolve, where dropping the `src` would otherwise persist a foreign signed URL. That is now pinned by its own test, which fails without the flag.

**3. The backfill command wrote a signed URL into the stored body.**
`$media->getUrl()` returns a 30-minute signed route on a private disk, so the migrated HTML carried a signature that expires 30 minutes after the queued job runs. It now writes `data-id` only, matching what the panel and the agent path store.

**4. The backfill could import a file type the allowlist forbids.**
`addMediaFromDisk(...)->toMediaCollection('attachments')` bypassed `StorePendingUpload`, so a legacy SVG on the public disk would become an attachment. It now checks mime and size up front, so the dry run reports honestly and one bad row cannot fail the job queued from the migration.

### `pending-uploads` and `attachments` were unregistered collections

Only `logo` was registered, so `FileAdder::guardAgainstDisallowedFileAdditions()` returned early for every upload this PR adds and the library-level mime gate never ran. Both collections are now registered with `acceptsMimeTypes(UploadAllowlist::mimeTypes())` on the workspace and the five record models. That is what closes defect 4 at the library layer rather than only in our action.

I did not remove the explicit size check from `StorePendingUpload`. `max_file_size` already gates it inside `toMediaCollection()`, but catching `FileIsTooBig` only changes which error message an agent sees and leaves a catch the earlier check makes unreachable.

### What I deliberately did not swap for a first-party helper

| Ours | First-party alternative | Why it stays |
|---|---|---|
| `decode()` | `addMediaFromBase64` | Its re-encode equality check rejects line-wrapped base64 that our `strict` decode accepts. I verified both behaviours |
| `fetch()` + `SsrfGuard::pinnedClient` | `addMediaFromUrl` + custom `media_downloader` | The `getTempFile` contract has no streaming ceiling; ours aborts mid-download at 10 MB |
| Signed PUT route | `Storage::buildTemporaryUploadUrlsUsing` | `Storage::fake('local')` drops the callback, so every faked upload test would throw. Worth revisiting on S3 |
| `RichContentAttachments` | `SpatieMediaLibraryFileAttachmentProvider` | Requires an existing record and a model attribute; custom-field values are neither |
| `ShowMediaController` | `Media::toResponse()` | Core emits `filename="…Str::ascii(…)"`, which drops RFC 5987 and empties non-latin names, the bug `4e6c9cec2` fixed |

Two things did simplify: `UploadClaims` and `DeleteWorkspace` now delete through `->each->delete()`.

### End-to-end verification

Walked against the running app on the redis queue with Horizon up, not the sync driver.

- MCP over JSON-RPC with a real token: `server/discover` reports `supportedVersions: ["2026-07-28"]`, `tools/list` returns 39 tools including both upload tools.
- All three upload legs: base64, `create-upload-url` → `PUT` (204) → `upload-file`, and a 3 MB base64 body.
- Agent creates a note from `suggested_markdown`, both files claim into `attachments` on the note with the right `custom_field_id`. Stored body carries no signature.
- Referencing a file another note already owns is refused: "This file is unavailable for Body."
- A second workspace can reference the uuid but the file never re-claims, and the API returns the link unsigned to them, so there is no cross-tenant signed-link leak.
- `GET /media/{uuid}`: 200 signed, 403 unsigned, 403 tampered, 403 expired, 404 unknown. Images inline under `default-src 'none'; sandbox`, PDFs as an attachment, original filename preserved including spaces.
- Pasting a PNG into the editor in a real browser creates the media row on submit, claims it, and stores ``.
- Refusals: SVG, localhost, `169.254.169.254`, plain http, two sources at once, no source, and `upload_id` path traversal all return a clean tool error.
- `app:purge-pending-uploads` no-ops on fresh rows, rejects `--hours=0` with exit 1, and deletes aged pending rows while leaving attachments alone. Deleting a workspace removes both the media rows and the files.
- Public pages: home, pricing, blog, docs, terms, privacy, self-hosted and the published comparison pages all 200. The 37 to 39 tool-count change matches the live server and is pinned by five tests.

Screenshot of the agent-uploaded image rendering through the signed route is at `.context/reviews/pr699/agent-upload-rendered.png` in my workspace. I have not attached it here; say the word and I will.

### Gates

```
pint --test --parallel   passed
rector --dry-run         0 changed files
phpstan                  0 errors
type coverage            100%
pest --parallel --no-tia 4923 tests, 4921 passed, 2 skipped
```

An earlier run reported 7 errors. All were Postgres deadlocks on `drop table` and missing-column inserts from a parallel run I had killed; a clean re-run is green.

### Two items for you

**1. `post_max_size` on the hosted MCP host.** A 10 MB base64 upload is a 14 MB request. Local Herd is at 5 MB, and the request fails as HTTP 200 with no JSON-RPC body, so a client cannot tell it failed. The self-hosted image is fine: `serversideup/php:8.5-fpm-nginx` ships `NGINX_CLIENT_MAX_BODY_SIZE=100M` and `PHP_POST_MAX_SIZE=100M`. The Forge host needs checking, and it is worth deciding whether an over-limit body should return a JSON-RPC error rather than an empty 200.

**2. A sibling branch adds a `chat-attachments` collection.** `6b2c12f0d feat(chat): accept a csv upload and hand large files to the wizard` writes a collection that is not in `App\Enums\MediaCollection` and not covered by `.ai/rules/file-uploads.md`, which states that every durable file uses a named collection from that enum with exactly two exemptions. Whichever branch merges second should reconcile that.

### Smaller notes, none blocking

- The editor signs attachment URLs at hydrate for 30 minutes (`MediaUrlGenerator::SIGNED_URL_MINUTES`). A form left open past that window holds links that will 403 if the browser refetches them. Pre-existing Filament behaviour for any private-visibility editor, worth knowing for a long-lived note form.
- `upload-file` rejects a `data:image/png;base64,…` payload, which is a shape agents do send. medialibrary strips that prefix; we do not. Two lines in `decode()` would close it, and I left it out because it is outside what you approved.
- The first panel save of an agent-created body normalises TipTap whitespace, so it logs one activity entry. Every save after that is byte-identical, which I confirmed in the browser.

ManukMinasyan:
## Pre-merge review, `ae353077b`

**Verdict: needs-human.** Gate is green, the upload lifecycle works end to end against the real MCP server and the real browser, and five fixes landed on the branch during the review. Five items need your call before merge.

### Verified live, not from tests

Real MCP protocol against the running app, Sanctum token, Redis queue, Horizon running from this checkout.

| Journey | Result |
|---|---|
| `create-upload-url` to signed `PUT` to `upload-file` | 204, then a media uuid with the client filename preserved |
| `base64` and `upload_id` sources | both store, both keep the original name including spaces |
| Stored body after an MCP write | **unsigned** stable `/media/{uuid}`, no expiry to rot |
| Claim on write | row moves `workspace`/`pending-uploads` to `note`/`attachments`, uuid unchanged |
| Panel render | signed 30 minute link, filename shown |
| **Panel save of a body rendered with a signed URL** | storage still unsigned. This is the regression the PR exists to fix, and it holds |
| Clear the body | media row, disk file and URL all gone (1 to 0, then 404) |
| Media route negatives | unsigned 403, tampered 403, unknown uuid 404; `nosniff`, CSP sandbox and attachment disposition all present |
| MCP auth and discovery | 401 with `WWW-Authenticate` + `resource_metadata`; `GET /mcp` 405, never 200 |
| Public pages | home, pricing, press, terms, privacy, blog, compare all 200; `/developers/mcp` renders the Files section, no YAML parse error |
| Tool count | 39 registered, and every marketing surface now says 39 |

### Fixed during review

1. **`CustomFieldSchema` returned HTTP 500 on a stored type with no enum case.** One legacy `markdown-editor` row broke `resources/read relaticle://schema/note` and leaked `"markdown-editor" is not a valid backing value for enum` to the agent. `CustomFieldType::from()` throws where the chat describer already used `tryFrom()`. Cold-reproduced, fixed, regression test added. (`5587a89e9`)
2. **The backfill aborted the whole run on one bad row.** `Artisan::queue()->afterCommit()` runs inline under `QUEUE_CONNECTION=sync`, and `transactionsManager->commit()` sits outside `transaction()`'s try, so the throw escaped `runMigration()` and `repository->log()` never ran. The migration then re-runs on every container start, which crash-loops the self-hosted image under `set -e`. (`c5ac2eac7`)
3. **SSRF allowlist gaps:** Teredo `2001::/32`, RFC 8215 local-use NAT64 `64:ff9b:1::/48` and ORCHIDv2 `2001:20::/28` all report public to `filter_var` and were missing, while their sibling translation prefixes were already covered. (`289131fd0`)
4. **Descriptor leak:** the `readStream` handle in `StoreAgentUpload::takeTemporary()` was never closed, leaking one fd per signed-PUT upload on a queue worker. (`cfb69af10`)
5. **`media.custom_field_id` dropped; attachments now belong to the record.** With one rich-editor field per entity today the column discriminated nothing, while scoping the release to a single field meant a release bug would delete a sibling field's files. Note the subtlety: `->where('custom_field_id', ...)` was also what kept the release away from `logo` rows, so the replacement is explicitly scoped to the `attachments` collection. Two guard tests (a second field's attachments survive; a company logo survives) were written against the old code first, so they pin behaviour rather than describe the new implementation. Side benefit: the same image can now appear in two fields and survives removal from one, which used to be an unsatisfiable validation error. (`ae353077b`)

### Needs your call

1. **`create-upload-url` has no per-workspace budget.** The 60/hour limiter is only on `upload-file`. The PUT receiver is `throttle:60,1` keyed by IP on an unauthenticated route, and `app:purge-pending-uploads` only sweeps `tmp/` after 24 hours, so any trial signup can park a lot of 10 MB bodies on the single Forge host uncounted. Both the budget and the retention are product decisions, so I did not change them. The 5 minute PUT expiry makes a 1 hour tmp cutoff generous.
2. **The receiver writes the body before enforcing the ceiling.** `Content-Length` is a client claim. The mutation audit found the pre-read guard is **untested in isolation**: deleting it keeps all 40 upload tests green, because the post-write check returns the same 413 the test named "before reading" asserts.
3. **`CustomFieldValue::save()` locks the owning record on every custom-field write**, not just rich editor, so a record with N custom fields pays N extra transactions plus morphTo and `FOR UPDATE` selects. Only a rich-editor value can hit the race it protects. No concurrency test pins the placement, so this wants a reproducing test before anything moves.
4. **Dead statement in the schema migration.** The `custom_properties->workspace_id` backfill matches zero rows on every install: nothing ever wrote that shape outside this branch's own history, and the spec says logo rows keep `workspace_id` null. Three independent passes flagged it. Also `->after('id')` is a no-op on pgsql.
5. **Four upload facts are re-spelled as prose.** The 10 MB ceiling, the extension list, the 60/hour limit and the 30 minute expiry are hardcoded in `#[Description]` attributes and lang strings while each already has an owner. `CreateUploadUrlTool:42` already interpolates the owner correctly one line over.

### Gate

`pint` pass, `rector` 0 changed / 0 errors, `phpstan` 0 errors, type coverage 100.0% (relative paths, so not a stale worktree cache), `pest --parallel --no-tia` **4954 tests, 4952 passed, 2 skipped, 23861 assertions**. Re-run green after every fix above, including the final refactor.

Mutation audit, 4 branch hunks worst-first plus failing-first proofs for each fix. `TemporaryUploads::belongsToWorkspace`, the `MediaLookup` cross-workspace guard (which fails at both the MCP and panel surfaces when reverted) and the media 404 guard all go red. The receiver ceiling is the one gap, item 2. The cap means the branch's remaining hunks were not audited.

### Notes, not blockers

- "22 custom field types" on 7 public surfaces is wrong and **pre-existing**: the enum has 21 cases and `file-upload` plus `markdown-editor` are disabled, so 19 are usable. The disabled list is byte-identical to main, so this branch did not cause it.
- Self-hosted storage is safe: `compose.yml` mounts the named `storage` volume at `/var/www/html/storage/app`, so `uploads/` and `tmp/` survive `docker compose pull && up -d`. The shipped image allows 100M bodies, clearing the ~14 MB a 10 MB base64 payload needs.
- The Forge host's `client_max_body_size` and `post_max_size` are not in the repo and remain the one deployment item I could not verify.
- Scope: the API owner, chat include and schema owner commits are required by the new `SurfaceParityTest`, so they are not drift.
- Dropping `media.custom_field_id` means deleting a rich-editor field now releases its images on the next save of any rich-editor value on that record, instead of stranding them. That closes a deferred follow-up, but the deletion is no longer undone by restoring the field.

<sub>fp-sha:ae353077b</sub>

PR #735, opened 2026-09-12
feat(onboarding): shape the workspace and open it on a setup conversation

ManukMinasyan:
## Summary

Signup wizard and workspace shaping
- Rebuild the signup wizard into three steps: workspace, attribution, use case. The invite step is gone, and the workspace name and handle start empty.
- A handle the user did not type is the model's to settle during the save, so two signups racing for the same default no longer die on a unique error. A guarded migration adds the unique index on the workspace handle that production was missing.
- Use case sub-options are a required multi-select on one axis per use case. Other takes a free-text answer in `onboarding_other_use_case`, and the sysadmin workspace view and table show it, searchable.
- Opportunity stages come from a per-use-case preset at workspace creation, with a customer success sample data set that matches the presets.
- The assistant gets the use case, context, stage names, Other text and a referral line in an `<onboarding>` prompt block. The tool schema wins when a legacy workspace's stages differ.

Setup conversation
- On by default. `RELATICLE_FEATURE_SETUP_CONVERSATION=false` turns it off. The test suite runs with it off and the onboarding tests opt back in.
- Every new workspace gets one setup conversation, marked by a `purpose` column on `agent_conversations`. Signup lands on it and the assistant speaks first, as a live streamed turn the owner watches arrive. The prompt that turn runs on never shows in the transcript, and a turn that dies before writing no longer leaves it there as the user's words or counts as the user having talked to Rela.
- Inside the setup thread the assistant runs in a create-only setup mode: update, delete and schema tools are absent, pasted contacts get a proposal on the first reply, and a change or delete request links the record page instead of saying unsupported.
- Setup proposals wait seven days instead of one (`setup_pending_action_expiry_minutes`).

File door
- The composer takes one CSV or TXT file (10 MB). Up to 25 rows are inlined into the user turn as fenced data. A larger file gets a stored reply with Import as people and Import as companies links that open the wizard on the mapping step with the file already loaded.
- Attachments are medialibrary rows in the workspace's `chat-attachments` collection, named by the `MediaCollection` enum and scoped on the row's own `workspace_id`. They are single-use per message and pruned hourly after 24 hours. The collection pins the `local` disk on purpose: both CSV readers take a filesystem path, and `Media::getPath()` is only a real path on a local-driver disk.
- The import wizard's parse-and-store logic moved into `ImportFileLoader`, shared by the upload step and the chat handoff.
- The attach button leads the composer's control row on its own. The model picker sits with the counter, mic and send on the trailing side, the arrangement most chat products share. The homepage hero twin mirrors it.

Sample data
- Owners can remove the seeded sample records from the activation checklist once the workspace has a record of its own.

Design spec and plan are in `docs/superpowers/`. The CI browser job now uploads its screenshots and Laravel log when it fails.

## Deployment note

Every upload path assumes one filesystem shared by web requests and queue workers. That holds today because production is a single Forge host where web and Horizon share the filesystem, and the self-hosted `compose.yml` mounts one `storage` volume into the app, Horizon and scheduler containers. It does not hold on Laravel Cloud, where the filesystem is ephemeral and not shared between replicas or with worker clusters.

Introduced here:
- The `chat-attachments` collection on `local`, read through `ChatAttachment::absolutePath()` by the inline reader and the wizard handoff.

Already on main:
- The import wizard keeps each import in a SQLite file under `storage/app/imports`, written by web requests and read by the three `imports` queue jobs. This is the blocker: no bucket fixes it, the store has to move to Postgres.
- Workspace and company logos pin the `public` local disk.
- The MCP upload flow writes a temp file in one request and reads it in a later one.
- Livewire temporary uploads use the default local disk.

Nothing in this PR changes the Cloud picture, and there is no useful partial fix. A configurable disk without a stream-to-temp reader would let `MEDIA_DISK=s3` break attachments outright, and a full media fix still leaves the import store dead. The object storage work is one branch of its own, import store first.

## Test plan
- [x] `php -d memory_limit=2G vendor/bin/pest --parallel --no-tia` (5057 passed, 2 skipped, on the final tree)
- [x] `vendor/bin/pest --type-coverage --min=100` (100.0 %)
- [x] `vendor/bin/phpstan analyse` (0 errors), `vendor/bin/rector --dry-run` (no changes), `composer test:lint` (pass), all on the final tree
- [x] `vendor/bin/pest tests/Browser/Onboarding/SetupConversationBrowserTest.php tests/Browser/Chat/ComposerAttachmentTest.php` (3 passed)
- [x] Real-stack walk with Horizon, Redis and Reverb: register as Recruiting, land on the setup thread and watch the greeting stream in (light and dark), pasted candidates to one proposal card, approve, a delete request answered with a record link, a 5-row CSV to a proposal, a 40-row CSV to the handoff and both mapping steps, remove sample data from the checklist, a second workspace opens without a setup thread
- [x] After the composer control row change: `vendor/bin/pest tests/Browser/Chat/ComposerAttachmentTest.php` (1 passed on the final tree), dashboard composer at desktop and 390px, and the homepage hero twin

Follow-up: #788 collects the import research and the options for letting Rela drive the import wizard.


ManukMinasyan:
## finalize-pr review

Gate green, sixteen journeys exercised against a production-shaped stack, six commits applied on top of the branch. Five questions below are yours, not mine.

Verdict: **needs-human**, on the two counts the verdict script pins: five open asks, and a mutation audit that found five production hunks no test covers.

### Found and fixed

**1. A wide attachment put 9.6 MB into the prompt** (high, `packages/Chat/src/Support/AttachedRows.php`). The row cap bounds height, not width. A 9.6 MB CSV of 25 rows and 2,000 columns passes `ImportFileLoader::inspect()`, so it takes the inline path, and `block()` assembles **9.58 MB** of text, roughly 2.5 million tokens. Measured on a generated file, not estimated. Typed text is capped at 5,000 characters (`ChatController.php:117`); the block was appended at line 163, after that check, with no ceiling of its own. The spec says a file of 25 rows or fewer is treated "the same as a paste", and it was not. That message is a constructor arg on `ProcessChatMessage`, so it is serialised into the Redis payload, stored in `agent_conversation_messages.content`, and re-sent on every later turn in the conversation. One reserved credit buys a request no provider will accept. `AttachedRows` now owns the whole inline decision and a file over the ceiling takes the wizard door that already exists for files over the row cap. Pinned by `it('hands off a file that is wide rather than tall, instead of inlining megabytes of prompt')`, which returns `processing` instead of `stored` without the fix.

**2. A failed opening turn made the setup thread permanent** (high, `packages/Chat/src/Jobs/ProcessChatMessage.php`). `failed()` called `persistFailedTurn()` unconditionally, writing a continuation user row plus a visible "encountered an error" assistant row. `StartSetupGreeting`'s guard is an unfiltered `$conversation->messages()->exists()`, so both rows disabled the greeting permanently. Every signup whose first model call failed landed on a thread whose only words were an error, and was never greeted. This contradicts spec 5.3 and the action's own docblock at `StartSetupGreeting.php:36-40`. The greeting is the only continuation dispatched without a `resumesTurnId`, which is the discriminator used. Reverting the fix writes 2 rows where the test demands 0.

To be precise about what "retries" now means: the thread is left empty, so the next open re-greets **once the 180 second `Cache::add` lock from the dead attempt expires**. Inside that window the owner sees an empty thread and can type into it.

**3. The hourly purge threw as soon as two attachments were due** (med, `packages/Chat/src/Commands/PurgeUnsentAttachmentsCommand.php`). `$media->model` inside the loop is a lazy load, and `AppServiceProvider` sets `preventLazyLoading(! isProduction())`. Eloquent arms that guard only when a query hydrates two or more rows, so the command works on one attachment and raises `LazyLoadingViolationException` on two, in local, CI and staging. Production is unaffected, which is exactly why it would not have surfaced. The existing test swept exactly one row. Now eager-loaded and chunked with `chunkById`, pinned by a two-row sweep.

**4. The media widen had no lock timeout** (med, `database/migrations/2026_09_16_100000_...`). `bpchar` to `varchar` is a function cast, not binary coercible, so the column rewrites the heap and rebuilds the polymorphic index under ACCESS EXCLUSIVE. Production `media` is 9,266 rows, so the rewrite itself is about 50 ms. The risk is the unbounded wait: one in-flight Horizon media write makes the ALTER queue, and every read of `media` then queues behind its pending lock. The sibling migration in this branch already sets `SET LOCAL lock_timeout = '5s'`.

**5. Sample removal was authorised only through the action.** All three negative cases called `RemoveSampleData::execute()` directly, which the project's own testing rules call an isolated unit test of an action. Nothing called `removeSampleData` through the Livewire component as a non-owner, so the component could lose its workspace scoping and stay green. Added that case; it asserts 403 and that no sample record was removed.

**6. A cleanup pass over the five fixes** (`72205043e`). The byte-versus-character one matters: the inline cap was written with `mb_strlen`, so it counted characters while the queue payload and the `content` column meter bytes, and a 65,536-character CJK block cleared it at roughly 196 KB. Now `strlen`. The rest is test hygiene: the two-attachment purge case folded into the test that already covers the sent and fresh ones, the five-model sample assertion collapsed from four copies into one helper, and a loose `->throws(HttpException::class)` member test dropped as the weaker duplicate of the component-level 403.

### Yours to decide

1. **Deploy ordering.** The flag defaults to `true` in `config/relaticle.php`, and `CreateSetupConversationListener` is a synchronous listener that writes `purpose`. If any request is served by new code before `migrate` finishes, that write fails with `42703` and the signup 500s; the wizard's `onboarding_other_use_case` write and a 36-character conversation id into a still-26-character `media.model_id` fail the same way. Two things soften this and neither settles it: the deploy fires on a published release rather than on merge, and the Forge deploy script is not in this repo, so **I could not check whether it runs `artisan down`**. If it does, there is no window and nothing to do. If it does not, either add maintenance mode around `migrate` or ship this release with `RELATICLE_FEATURE_SETUP_CONVERSATION=false` and flip it once migrations land. Your call which.
2. **Remove sample data takes edited samples too.** The filter is `creation_source = system`, and nothing resets that on update. A seeded company the user renamed, re-staged and attached their own notes to still goes. The confirm reads "Delete every sample record? Your own records stay." For someone who worked inside the demo data, which is the intended first-run path, that sentence is not true.
3. **Removal is a soft delete.** Spec 5.7 says records go "through the models so custom field values and pivots follow". The custom-fields `deleting` hook returns early unless `isForceDeleting()`, so `custom_field_values` survive for every removed sample. Verified in the browser either side: 21 sample rows moved to trash, the three own records untouched, and `hasSampleData` then reads false so there is no path back. The user-visible result is right; the spec sentence is wrong. Correct one or the other.
4. **The slug unique index has no dedupe.** Production is safe and I checked: 3,608 workspaces, all slugs distinct, and `pg_indexes` confirms the index is genuinely missing there. The residual case is a self-hosted install that lost the index the same way and then accumulated a duplicate. `migrate` throws, and under `set -e` that crash-loops the container and takes Horizon with it. The 2026_02_11 migration already carries a suffix loop that could be reused.
5. **The flag defaults on with no provider check.** `AiModelResolver::autoPick()` returns `$chain[0]` even when no descriptor is available, and `StartSetupGreeting` has no availability guard. A self-hosted install with no provider key gives every new owner a first screen whose opening turn fails.

### Mutation audit

Twelve hunks, worst-first over the slice-2 production code. Seven red, five gaps:

| Hunk | Result |
|---|---|
| `ChatAttachment::find` morph `ownedBy` | red |
| `RemoveSampleData` owner check | red |
| `CreateSetupConversation` personal-only guard | red |
| `resolveAttachment` single-use check | red |
| `PendingActionService` setup expiry | red |
| `CrmAssistant` setup-mode tool exclusion | red |
| `EnsureHostedWorkspaceAccess` import route | red |
| `ChatAttachment::find` `workspace_id` filter | **green**, redundant with `ownedBy`, which is pinned |
| `ImportAttachment` `can('create')` | **green**, a viewer following the handoff link is unpinned |
| `ImportAttachment` `fileExists()` 410 | **green**, the purge running between reply and click is unpinned |
| `StartSetupGreeting` empty-thread guard | **green**, shadowed in tests by the `turnInFlight` early return |
| `StartSetupGreeting` greet-once `Cache::add` lock | **green**, same shadow; delete the lock and nothing goes red |

The two `ImportAttachment` gaps are real authorisation and state paths, not redundancy. The two `StartSetupGreeting` gaps matter more now that a failed greeting legitimately re-greets.

### Reported, not touched

- **The purge still carries an N+1.** `messages()->exists()` fires once per row inside the chunk. The clean shape is two phases: keep the chunked media delete, then one sweep for conversations left with no messages and no attachments. Worth doing when the volume justifies it; today the set is whatever went unsent in 24 hours.
- **"A message the user can see" now has four owners:** `TranscriptScope::apply()`, `WorkspaceActivationFacts::hasUserChatMessage()`, the raw `messages()->exists()` in `StartSetupGreeting:68`, and `opensTheThread()`. Drift rather than a live race, but it is the shape the one-fact-one-owner rule names.
- **`AttachedRows::block()` assembles the whole block before the cap measures it,** so the wide file it guards against still costs roughly 104,000 sanitise calls before being refused. The real bound is the missing upload quota below, which is why that is the better place to spend the fix.

- **The setup-mode tool filter is a denylist** (`CrmAssistant.php:941`). A new Update or Delete tool added to `toolClasses()` later is included in setup mode by default, and the guard test asserts a count difference that grows on both sides, so it cannot catch it. An allowlist, or a capability the tool declares, would close it.
- **Single-use is check-then-set.** `resolveAttachment` reads `! isSent()` with no lock, and `markAttachmentSent` runs after `dispatch`. Two concurrent sends of one attachment both pass and both reserve a credit. `ImportAttachment` already shows the right shape with `lockForUpdate()`.
- **`ActivationChecklist` redirects to the `Referer` header.** Not exploitable: the Livewire POST is same-origin and CSRF-gated. It is also load-bearing, because the checklist sits in the sidebar footer on every page, so redirecting to the dashboard unconditionally would regress the UX. Validate the host rather than dropping the header.
- **The upload route has no per-user quota.** Thirty uploads a minute at 10 MB each, purged only after 24 hours, on the same volume as the app.
- **`InviteWorkspaceMemberTool` survives setup mode.** It is the one tool left there whose approval sends outbound email. The spec lists neither its inclusion nor its exclusion.

### Verification

`pint`, `composer test:lint`, `rector --dry-run` (0 changed), `phpstan` (0 errors), type coverage **100.0 %**, and `pest --parallel --no-tia`: **5,067 tests, 5,065 passed, 2 skipped, 0 failed**. CI is green on every check at the branch SHA this review started from.

Sixteen journeys were exercised. Eight carry an annotated screenshot, read back before it was hosted; the other eight were verified by reading the DOM, the database, or a test, and each is recorded in the bundle ledger with the shot it leans on. The greeting is a live model turn, so its wording is not deterministic; what is pinned is that it names the workspace's own first and last stage, which it did.

### Journeys

**J2, the wizard opens empty.** Three steps, "Your name" prefilled for a first-run user, workspace name and handle carrying only a placeholder.

step 1

**J7, signup lands on the setup thread and the assistant speaks first.** Registered as Recruiting; the greeting streamed in naming recruiting and this workspace's own pipeline ends, Sourced through Hired.

greeting

**J8, the first reply to a paste is a proposal.** Three candidates pasted as loose columns, no clarifying question, each field mapped. Approving resolves the batch per item, "1 of 3 resolved", with a link to each created record.

proposal

**J11, a 40-row CSV takes the wizard door.** No model run, no credit spent, the file rendered as a card and not an empty bubble.

handoff

Following Import as people created one `Import` at `status=mapping` carrying all four headers and 40 rows, and landed on the mapping step. Clicking it a second time reused that import rather than creating a second.

**J10, sample data goes once own records exist.** Counted either side: 21 sample rows removed, the three pasted people untouched.

remove sample data

**J14, the Other free text is readable.** A second signup answering Other with "Grant applications and funder relationships" shows it in the sysadmin workspace view, with the sub-options rendering as the empty glyph because Other stores none.

sysadmin

**J1, the homepage hero twin.** Attach leads on its own, model picker, mic and send trail. Light and dark, no horizontal overflow.

hero light

hero dark

_Assessed: correctness, over-engineering, Laravel and Spatie conventions, project guidelines, appsec, functional behaviour in a real browser, operability and scale (Tier 3)._
_Not assessed: load and abuse at scale; the day-2 nudge, exit survey and SystemAdmin funnel (slice 3, no production code here); a live cross-provider failover; and the wording of any given model turn._

fp-sha:72205043e

ManukMinasyan:
@copilot resolve the merge conflicts in this pull request

PR #784, opened 2026-09-15
Impersonate a customer from the sysadmin panel

ManukMinasyan:
Adds impersonation so support can reproduce a tenant-specific bug by signing in as the customer, from the user list, the user page, or a workspace page (which lands directly in that workspace).

**How the handoff works.** Staff and customer sessions are isolated per host (#801), so no request can hold both identities. The panel action mints a relative-signed, single-use link valid for 60 seconds and addresses it to the app host. The link carries the target and the administrator. The app host resolves the administrator through the `sysadmin` guard's provider, re-checks the `impersonate` gate, claims the nonce, and only then assumes the customer. The same link returns 404 on the staff host and 403 on a second use. One code path serves path-routed and domain-routed installs.

**The session swap is deliberately quiet.** Firing `Login` would write `last_login_at`, which feeds the engagement buckets and the Mailcoach subscriber tags, and `Logout` cycles the remember token, signing the customer out of every device. Every sign-out path (panel menu, Fortify, invitation account switch, deletion interstitial) ends the impersonation instead of signing out.

**Identity proofs do not cross the swap.** An identity confirmation is session-scoped, so the swap clears identity proofs in both directions and re-stamps MFA completion, as `CompleteAuthentication` does at login. Without this, a confirmation the administrator made on their own account passed the provider-link gate as the customer. It also lets an MFA-enrolled customer be impersonated instead of bouncing to their two-factor challenge.

**Audit.** Start and stop are activity-log rows caused by the administrator, taken from the link and the session rather than a live guard. Every write made while impersonating carries `properties->impersonated_by`, with the causer left as the customer because the record is theirs.

Restricted to a SuperAdministrator, because `canManageCustomerAccess()` already withholds customer-record edits from an Administrator and impersonation grants exactly those edits. Minting requires a staff session that passed the second factor (#789).

Also fixes `ValidateSignature` running a signed route twice when the route itself throws `InvalidSignatureException` (own commit, own test).

**Verified.** Two-host tests in the session-isolation harness mint through the real panel action on the staff host and consume on the customer host with separate cookie jars. Single-host tests cover replay, expiry, a deleted administrator, a role that may not impersonate, the identity-proof reset in both directions, MFA customers, and every sign-out path. Walked in the browser: impersonate from the user page and the list, banner in light, dark and mobile, panel Sign out (customer remember token unchanged), banner Stop, app session gone with the staff session intact, audit rows caused by the administrator. Arch tests, PHPStan, pint, rector and 100% type coverage pass.

Reviewers need `pnpm run build` on this branch, or the banner's Tailwind classes are missing and dark mode renders unreadable.


ManukMinasyan:
## Final review: impersonation (`f27ff3dc8`)

**Verdict: needs-human.** No blocking bug found. The handoff, replay protection, audit trail and session swap all hold, and all four product questions are decided (below).

Reasons (computed):
- J6 (Administrator role sees no Impersonate button) not walked in the browser: the seeded Administrator is forced into MFA enrollment first. Covered by new panel tests instead.
- One fix reverted after failing the gate (details under Fixes)

## Decisions

The four open questions were decided by the author.

**Q1, credentials minted while impersonating: left as is.** Reproduced: while impersonating, Settings > Access Tokens created a real read token for the customer (expires Mar 16, 2027) with no activity row. By code read, `POST /oauth/authorize` and workspace invitations are open the same way. Accepted because impersonation is SuperAdministrator-only and start and stop are audited. Password, email, passkeys, MFA, provider links and account deletion stay closed: they need a proof only the customer holds.

token minted while impersonating
_Token field blurred before capture; the probe token was deleted afterwards._

**Q2, impersonation rewrites `current_workspace_id`: accepted.** Reproduced: browsing acme-sales as the customer saved it as their current workspace. Not saving it looked cheap but is unsafe: `ProcessChatMessage` reloads the user from the database and chat tools scope by `$user->currentWorkspace` (`packages/Chat/src/Tools/BaseWriteDeleteTool.php:73`), so chat used while impersonating would act on a different workspace than the one on screen.

**Q3, queued writes not tagged: fixed in `f27ff3dc8`.** `Context::dehydrating` adds the administrator id as hidden context when a job is queued during impersonation, and the activity hook falls back to `Context::getHidden('impersonated_by')` in the worker. Hidden context stays out of log output, and the scoped repository resets per job, so it cannot leak into the next job. The test queues a real job on the database queue, drops the session and runs `queue:work --once`; it failed before the fix with `null` and passes after. Not walked as a browser CSV import.

**Q4, owner-level destructive actions: accepted.** By code read, not walked: delete workspace, the Stripe billing portal and member removal stay reachable while impersonating, consistent with Q1.

## Fixes applied

| Commit | Change |
|---|---|
| `8640f045e` | Tests: Impersonate / Impersonate owner hidden for an Administrator and for an ownerless workspace; minted links carry administrator, nonce, expiry, signature and land in the viewed workspace. None of this was covered: reverting the `authorize` closures stayed green. |
| `31505b9a1` | `->authorize('impersonate')` / `->authorize('impersonateOwner')` through the existing policies, as `SubscriptionResource` does (Filament unshifts the record, `vendor/filament/actions/src/Concerns/CanBeAuthorized.php:80`). Removed the `getSysadminUrl` macro for `route('filament.sysadmin.resources.users.index')`, which already resolves the staff domain (two-host test asserts `https://staff.example.test/users`). |
| `5db1d26ae` | `abort_if($administrator === null)`; trimmed two comments over the 2-line cap and one that argued the diff. |
| `6ff2b0d5d` | Reverted my own edit that made the deletion interstitial call `StopImpersonationController`: the arch preset forbids controllers outside HTTP (full suite, 1 failure). The interstitial now stops inline. |
| `804728cf0` | Tests: the password-hash rewrite in `Impersonator::assume()` had no covering test. Factories share one memoized hash, so every user looked alike to `AuthenticateSession`; the new test uses a distinct admin password. Plus a banner test (shown only while impersonating). |
| `bbfb9900a`, `086f177d9` | `to_route()` per rector; `mutates(Impersonate::class)` on both resource test files. |
| `f27ff3dc8` | Q3: writes from jobs queued during impersonation carry `impersonated_by` (Laravel `Context`), with a real queue-worker test. |

Simplification: net -26 production lines across the review (27 added, 53 deleted, including the Q3 fix). Lean already after that.

No security fix was auto-applied. One was drafted and dropped: switching the swap to `session()->invalidate()` (as `CompleteAuthentication` does) would sign the admin out of sysadmin on path-routed installs, where staff and app share one session. Reported below instead.

## Verified end to end

**Two-host, prod-shaped** (`APP_PANEL_DOMAIN=app.brazzaville.test`, `SYSADMIN_DOMAIN=sysadmin.brazzaville.test`, `SESSION_DOMAIN` on the parent domain like prod). Staff login went through the real TOTP challenge.

| Journey | Result |
|---|---|
| J1 start from the user page | Confirm modal on staff host, link minted to `app.`, landed `https://app.brazzaville.test/acme-sales` with banner (light, dark) |
| J2 stop / sign out | Banner Stop returns to `https://sysadmin.brazzaville.test/users`, staff session kept, app redirects to login, 0 customer sessions left. Panel Sign out (path-routed) also ends the impersonation. `remember_token` and `last_login_at` unchanged in both. |
| J3 audit | Start and stop rows in sysadmin Activity, causer = administrator |
| J4 escalation | Q1 above |
| J5 sad paths | Used link: 403 on `app.`, 403 on apex, 404 on staff host. Tampered nonce 403, unsigned 403, DELETE stop without CSRF 419 |
| J7 impersonate owner | Landed in the viewed workspace, not the customer's current one (path-routed) |
| J8 mobile 390px | Banner 29px, message truncates, Stop stays at x=378 (path-routed) |
| J6 Administrator role | Not walked (MFA enrollment wall); covered by `UserResourceTest` / `WorkspaceResourceTest` |

Path-routed (the self-hosted default) was walked too: sysadmin stays signed in after start and stop on the shared host. The banner lives outside the scrolling `.fi-layout`, so the sticky topbar never slides under it.

confirm modal
banner light
banner dark
audit rows

**Production checks (read-only):** `CACHE_STORE=redis`, so the nonce `Cache::add` is atomic across workers. No workspace slug starts with `impersonate`, so the new reserved slug shadows nobody.

## Gate

| Check | Result |
|---|---|
| `composer test:lint`, `test:refactor`, `test:types` at `f27ff3dc8` | pass |
| `composer test:type-coverage` | 100% (relative paths, cache cleared) |
| `composer test:pest:full` (no TIA) at `f27ff3dc8` | 5191 passed, 2 skipped |
| CI on `5ed2960a7` | all green |

**Mutation audit: 10 of 10 production hunks go red when reverted** (gate check, nonce claim, landing membership, password-hash rewrite, previous-user restore, `impersonated_by` tag, logout interception, banner hook, reserved slug, identity-proof reset). Two were green before this review and got tests in `804728cf0`. Not audited (cap of 10): `ValidateSignature` (has its own test), `canImpersonate()` (reached by the gate-check and panel tests).

## Reported, no action

- Low: session keys cross identities on the swap. A test showed the admin's own `url.intended` and `mcp.oauth.workspace_id` surviving into the customer session. `invalidate()` is unsafe here (shared session on path-routed installs); forgetting known keys is the option if it matters.
- Low: an anonymous `DELETE /impersonate` with a valid CSRF token 302s to the staff host URL. That hostname is already public in the certificate SAN.
- Low: while impersonating, `sessions.user_id` is the customer, so their Browser Sessions list can show the support session's IP and browser.

_Assessed: correctness, over-engineering, conventions, project guidelines, appsec, functional behaviour (small-N browser), operability (prod cache store, slug collision, session domain, host exposure)._
_Not assessed: load and scale, abuse and rate limiting on the start route, deploy and migration lock (no migration in this PR)._

fp-sha:f27ff3dc8

PR #789, opened 2026-09-16
Require sysadmin second factor and add staff passkeys

### LINKED ISSUE #787: Require a second factor for sysadmin panel sign-in
The sysadmin panel authenticates `SystemAdministrator` with a password only: the model has no two-factor or passkey support and `SystemAdminPanelProvider` configures no `multiFactorAuthentication()`.

Impersonation (#784) makes that single factor load-bearing. A `SuperAdministrator` password now yields a full authenticated session as any customer, including customers who enrolled MFA on their own accounts. Before impersonation, the same password only reached customer metadata inside the admin panel.

**Scope**
- Add Filament's app (TOTP) authentication to the sysadmin panel and mark it required, so an unenrolled administrator is prompted to enrol at next sign-in.
- `SystemAdministrator` implements `HasAppAuthentication` and `HasAppAuthenticationRecovery`; migration for the secret and recovery codes.
- Recovery codes shown once at enrolment, regenerable from the sysadmin profile page.
- Passkeys for administrators: registration and management on the sysadmin profile page, and passkey sign-in on the sysadmin login, reusing the app panel's WebAuthn setup where the guard allows.
- Sweep `packages/SystemAdmin` for anything that assumes password-only login (it is excluded from PHPStan).

**Out of scope**
- Changing which roles may impersonate; that stays `SuperAdministrator` only.

**Done when** a `SystemAdministrator` cannot reach any sysadmin route, including the impersonation link, without a second factor, and the existing sysadmin feature tests pass with an enrolled administrator.

ManukMinasyan:
Require authenticator enrollment before administrators access protected sysadmin routes, with encrypted secrets and recovery codes.
Add staff passkey sign-in and profile management, with separate credentials, host binding, and authenticator confirmation before adding or removing passkeys.

A passkey is a complete sign-in on its own, not a second step after the password: a resident credential with user verification is already multi-factor, so answering the passkey prompt does not also prompt for an authenticator code. Enrollment still requires the authenticator app, and every route stays behind the enrollment check.

Staff passkeys are offered only where `SYSADMIN_DOMAIN` gives the panel its own host. Sharing a host with the customer panel means both share a WebAuthn relying party, so the sign-in picker would offer customer credentials that the staff endpoint must reject; the affordance is hidden in that case rather than failing on use.

Validation: the full test suite, Pint, Rector, PHPStan, 100% type coverage, and browser checks in light, dark, and mobile layouts.

Closes #787.


ManukMinasyan:
## Review: everything found is now fixed, two test gaps disclosed

Updated after the follow-up commits. The requirement in #787 is met and proven in a real
browser: an administrator holding a correct password reaches **no** sysadmin route without
an enrolled second factor. I walked all six privileged routes plus the passkey registration
endpoint as an unenrolled admin; every one redirects to enrolment.



Five issues came out of the review. All five are fixed on the branch. What remains is two
mutation-audit gaps that no feature test can close, one architecture note, and one product
consequence worth being deliberate about.

| # | Issue | Outcome |
|---|---|---|
| 1 | An authenticator code could be replayed, and guessed without limit | fixed, `19a05530a` |
| 2 | Staff and customer passkeys collide when both panels share a host | fixed, `20f0f1af1` |
| 3 | "Turn off" offered for a factor the panel requires | fixed, `20f0f1af1` |
| 4 | A passkey signs in without the authenticator code, undocumented | documented in the PR body |
| 5 | Forged staff assertions were not converted into a clean rejection | fixed, `37de7382d` |

### 1. An authenticator code could be replayed, and guessed without limit

`EditProfile::appAuthenticationCodeRule()` called `verifyCode($value, $secret)` with
`shouldPreventCodeReuse` left at its `false` default, and neither passkey action had an
attempt limit. Filament's own three sensitive-action rules
(`DisableAppAuthenticationAction:61`, `SetUpAppAuthenticationAction:109`,
`RegenerateAppAuthenticationRecoveryCodesAction:60`) all pass `shouldPreventCodeReuse: true`,
keep a per-user `RateLimiter` at 5 attempts, and add `->rateLimit(5)`.

This mattered more than it looks: `AppAuthentication::$codeWindow` defaults to **8**, a
window of plus or minus four minutes, so roughly 17 codes out of a million were acceptable at
any moment against an endpoint that allowed unlimited tries. The attacker model is the one
this PR exists to defeat: someone holding a stolen sysadmin session who wants to plant a
durable credential.

Fixed by mirroring Filament's rule exactly, with two failing-first tests. Confirmed in the
browser: a code already spent by the login challenge is now refused on "Add a passkey" with
no credential written, and a fresh code completes the ceremony.

### 2. Staff and customer passkeys collided when both panels share a host

`SYSADMIN_DOMAIN` and `APP_PANEL_DOMAIN` both default to unset, so the default deployment
shape served both panels from one host. Both then had the same WebAuthn relying party, and
`GenerateVerificationOptions` deliberately sends `allowCredentials: []`, so the browser
offered customer credentials on the staff login:

```
GET  /sysadmin/passkeys/login/options
  -> {"options":{"rpId":"lusaka.test","allowCredentials":[],"userVerification":"required"}}
POST /sysadmin/passkeys/login
  body: {"credential":{"id":"lvGRmPHLRP8X0nusnB4szkI04ODumPmiYqgtQl66wyQ", ...}}  <- CUSTOMER credential
  -> 422 {"message":"Passkey not recognized. It may have been removed from your account."}
```



The server was always right, and `VerifyPasskey`'s guarantee held: only
`system_administrator_passkeys` is searched, so a customer credential is never a staff
credential. The problem was that the feature was offered in a state where it fails, and the
error blamed credential removal.

`SystemAdministratorPasskey::hasDedicatedRelyingParty()` now compares the two relying
parties. Where they match, passkey sign-in and registration are hidden rather than left to
fail on use. The profile section still renders where credentials already exist, so they stay
removable after a panel moves back onto one host. `compose.yml` also now forwards
`SYSADMIN_DOMAIN` in all three services, which it never did before, so a self-hosted install
can actually reach the separated configuration.

### 3. "Turn off" was offered for a factor the panel requires

Filament keeps its disable action even under `isRequired: true`. Using it succeeded, then
every navigation redirected to enrolment and any registered passkey became unusable, because
passkey management sits behind the same middleware. A small `RequiredAppAuthentication`
subclass now drops that one action and keeps recovery-code regeneration.



### 4. A passkey signs in without the authenticator code

Neither Filament's `EnsureMultiFactorAuthenticationIsEnabled` nor the new
`RequireSecondFactor` records that a session *passed* a challenge; both check only enrolment.
`PasskeyLoginController::store` calls `$guard->login()` directly, so an enrolled
administrator signing in with a passkey gets a full session with no TOTP prompt. That is
defensible, since a resident credential with user verification is already multi-factor, but
it was not what the PR body said. The body now states it.

### 5. Forged staff assertions were not converted into a clean rejection

`app/Actions/Passkeys/VerifyPasskey` on `main` already wraps forged-assertion `Throwable`s
into `InvalidPasskeyException`, with a comment recording why: the bytes surface as PHP
warnings and bare `InvalidArgumentException`s, not only the documented `WebauthnException`.
The staff copy did not. It does now.

Worth noting honestly: **I could not reproduce a 500 through it.** My probe returned 422
because the payload failed an earlier guard. The customer-side hardening is the evidence, and
the wrap can only ever convert a 500 into a 422.

### What is still open

**Two production hunks no test covers.** The mutation audit ran 12 mutants over the branch's
own auth hunks and killed 10. The survivors are the `canAccessPanel()` gate in
`PasskeyLoginController:44` and the duplicate `credential_id` check in `StorePasskey:39`.
**untested because** both sit behind a real WebAuthn ceremony a feature test cannot produce,
and `VerifyPasskey` is `final readonly`, so it cannot be stubbed at the container seam. Both
were exercised in the browser; neither is guarded in CI.

Disclosed about the fix itself: the two rate limiters cover for each other, so removing
either alone leaves the tests green; removing both is caught. `shouldPreventCodeReuse: true`
is killed individually.

**The staff passkey actions still fork the vendor** where `app/Actions/Passkeys/` extends it,
with arch-test `ignoring()` entries already paying for that. Issue 5 is exactly the cost of
that fork: a hardening applied once had to be applied twice. Worth a follow-up, not a blocker.

**Staff passkeys now ship dark by default.** The gate in issue 2 means they are invisible on
any install that does not set `SYSADMIN_DOMAIN`, including local development. That is the
intended effect, and better than offering a feature that fails on use, but it is a real
narrowing of what this PR advertises and should be a conscious call.

### Verification

Eight journeys through a real browser, driven by a CDP virtual authenticator (`ctap2_1`,
resident key, user verification forced). Every WebAuthn ceremony is real.

| Journey | Result |
|---|---|
| Customer app panel: register a passkey, clear session, sign in with it | pass |
| Unenrolled admin against all six privileged routes plus `/passkeys/options` | pass, all redirect to enrolment |
| Enrol TOTP, recovery codes once, sign in with password + code, sign in with a recovery code | pass |
| Staff passkey register / sign in / remove, re-walked after the code-reuse fix | pass |
| Sysadmin login surface, light and dark | pass |
| Staff passkey sign-in with a customer passkey on the same host | reproduced, then fixed |
| Public and customer sweep, robots header, rate limit | pass |
| Post-fix profile and login surface on a single-host config | pass |

The only paying-customer surface this PR touches is `config/fortify.php`. Customer passkeys
still register and sign in after that rewrite: a resident credential was created on the
authenticator with `rpId` matching the app panel, the row landed in `passkeys`, and after
clearing the session that same credential signed the customer back in with `last_used_at`
written.



The staff sign-in affordance, captured while a dedicated relying party was configured. On a
single-host install it is now hidden entirely:





Gate, all green: pint, `test:lint`, `test:refactor`, PHPStan L7 (0 errors), type coverage
100%, and `composer test:pest:full` without TIA at **4992 tests, 4990 passed, 2 skipped, 0
failed**. `config:cache` and `route:cache` both succeed with the four new routes surviving.
Encryption at rest is real, not assumed: a 16-character TOTP secret persists as 228 bytes of
ciphertext and the recovery codes as 1108, which is what makes the migration's `text` columns
necessary. Rate limiting on the guest passkey endpoint was observed rejecting at the cap.

Four things checked that are **not** problems: the migration's explicit `index('user_id')` is
required, because PostgreSQL does not auto-index foreign-key columns; the
`config/fortify.php` closure is an IIFE, so `config:cache` stores its result;
`RequireSecondFactor` appears twice in the route middleware but `gatherMiddleware()` dedupes
it; and `@source` was added to the correct theme file.

Separately, `app/Listeners/LogPasskeyEvent.php` and its test were removed at the author's
request. It predated this branch and nothing else referenced it. Consequence, recorded
without argument: passkey registration, verification and deletion now emit no application log
line on either panel, so the only record of a staff passkey sign-in is `last_used_at`.

fp-sha:37de7382d

PR #345, opened 2026-06-13
feat(chat): proposal UX overhaul — per-record approval, docked editing, inline field edits

ManukMinasyan:
## Summary

Overhauls the AI chat write-proposal/approval UX, benchmarked against Attio's assistant, across six increments. All work is scoped to `packages/Chat` + tests. The result: proposals read clean (no duplicated data), you control each record in a batch individually, the proposal card docks at the composer while a decision is pending, and you can edit any field inline before approving.

> **Note (updated after business review):** the final commits deleted the continuation/resume backend and the REST inline-edit endpoints in favor of a Livewire-component design. The sections below describe the **shipped** behavior, not the intermediate steps.

### Phase 1 — de-noise & visual parity
- Agent prompts no longer enumerate proposed record data in prose (it used to render three times: a markdown table, the card, then a confirmation table). The card owns the data; the assistant replies in one sentence.
- Type-aware field rendering via a shared `_proposal-field` partial: links as links, choice fields as badge pills, booleans as Yes/No pills, multi-value links as separate anchors.
- "View" links labeled with the record name; calmer Create / Save changes / Delete + neutral Discard buttons; `⌘↵` approve on single proposals.
- Page header (H1 + tab title) syncs to the conversation title after the first turn, no reload.

### Phase 2a — per-record resolution
- Multi-record Create proposals resolve **record-by-record**: navigate with `‹ N / M ›` and Approve (Create) or Skip (Discard) the current record individually. There is **no** "Create all / Discard all" bulk button — each record is resolved on its own card view.
- Only the approved subset executes. Each item's outcome is reflected **inline on its own card** (Approved / Skipped) as it resolves — there is **no** trailing assistant summary sentence (the AI continuation was removed; see Fix below).
- A `approveItem`/`rejectItem` path in `PendingActionService` accumulates per-item outcomes in `result_data` (string-keyed `items` + running `ids`); the finalized `{ids, type, count}` shape is preserved. The whole-batch `approve()`/`reject()` stay atomic. Per-item writes are guarded by `lockForUpdate()` and are idempotent (an already-resolved index is a no-op). No schema change.

### Phase 2b — docked proposal + composer block
- While a proposal is pending, its card docks directly above the composer under a "Review before continuing" header, and the composer input is hidden (`x-show="!hasPendingProposal"`) so you must resolve the proposal before typing again — removing the previous silent-supersede behavior. On resolve, the card moves inline as the audit trail and the composer returns. The block is restored on reload while the proposal is still pending.
[...]

PR #326, opened 2026-06-11
feat(chat): stream identity, honest failures, batch proposals, resumable continuations

ManukMinasyan:
Fixes the production chat failures documented in issue #316's family and the June 10 production transcript: duplicated streamed text, duplicate proposal cards, lost multi-step plans, silent rate-limit failures, and the one-at-a-time approval grind.

## What changed

**Client stream identity (the root of the visual chaos).** Every laravel/ai stream event carries `invocation_id` (fresh per job attempt); bubbles now bind to it. A 429-retried job that re-streams the turn resets the partial bubble instead of appending a duplicate copy; trailing deltas of a still-open turn can never leak into a later bubble; `stream_end` finalizes the bubble it belongs to, not whatever is last.

**Honest failures, recoverable everywhere.** A failed turn shows the actual error in an amber banner with a Retry button (previously: garbage partial text and no error if the bubble was non-empty). Retrying a failed continuation hits the new idempotent `POST /chat/conversations/{id}/resume` endpoint; the chain-pause bubble gets a Continue button (was: "type continue"). Backoff windows show "Provider is busy — retrying (attempt N of 5)…" instead of silence. Stale failure banners clear once a later turn completes.

**Jobs stop lying about success.** Mid-stream SSE `error` frames threw nothing before — partial turns were persisted and billed as success with no `stream_end`. They now map to retryable/fatal exceptions like connection errors. Stream broadcasts are payload-disciplined: read-tool results are no longer broadcast at all (the client never rendered them; a 40-row list result blew Reverb's 10KB cap and killed the whole turn), proposal payloads drop the action-data echo, and an oversized event degrades to a reconcile-heal instead of killing the stream.

**Batch create proposals.** Create tools accept `records[]` (1–25, mirroring delete's `ids[]`): one grouped card, one approval, all-or-nothing execution in the approval transaction, per-record View links, batch-aware continuation labels. "Create 5 tasks" went from 6+ model calls / 5 approvals / ~49k input tokens to 2 calls / 1 click / ~13k.

**The model keeps the plot.** The `[approval]` continuation prompt is now prose with the record titles, internal ids, and optional plan progress ("Original request… 2 of 5 done"), ending the "the task 'Create Task' was created" parroting and the lost-plan stalls. Near-duplicate create proposals (same title within 15 minutes) carry an explicit warning on the card. Plan text is sanitized before being embedded in the privileged prompt.

## Verification

[...]

PR #803, opened 2026-09-17
feat(chat): record who authored a chat message as its origin

ManukMinasyan:
Chat messages the app writes itself (the setup greeting and the resume turn after a decision) were marked by a `meta->kind` flag. A reader that skipped the flag (`sentBy()`) tagged users `has-ai-usage` who had only seen the greeting.

Every message now carries an `origin` column owned by `MessageOrigin` (`typed`, `greeting`, `resume`). A synthetic row stores only a short opener. The turn's instructions travel in a `<turn>` block of `dynamicInstructions()`, so they are never stored or replayed.

Message filters are scopes on `AgentConversationMessage`: `typed()`, `withoutSynthetic()`, `ownedBy()` and `visibleTo()`. Conversation actions read through `AgentConversation::ownedBy()`, and `DB::table` readers keep their rows through `toBase()`. Conversation search no longer matches a synthetic opener.

Two migrations add the column and backfill legacy rows. They convert the three exact approval payloads, `continuation` rows, and the exact resume prompt that released versions stored unmarked when a turn failed. That last shape showed the full directive as a user message in the transcript.

SystemAdmin shows each message's origin. A new architecture rule keeps query filters on their model as scopes, and `tests/Arch/ConventionsTest.php` enforces it.

The same rule applies outside chat. Workspace membership was written three ways (`WorkspaceMemberSelect`, the API tenant scope, `TenantFkValidator`) and is now `User::memberOf()`. The two notification commands shared a copied local-hour query, now `User::atLocalHour()`.


ManukMinasyan:
## Final review: record who authored a chat message as its origin

**Verdict: approved.** No blocking bug was found. Review added one missing test (`c1afda8db`) and, after the author's decisions below, one SystemAdmin fix (`71bc3faa1`).

### Gate (head `71bc3faa1`)

| Check | Result |
|---|---|
| Pint, `composer test:lint` | pass |
| Rector (`test:refactor`) | pass, 0 changed files |
| PHPStan (`test:types`) | pass, 0 errors |
| Type coverage | 100.0% (fresh cache, relative paths) |
| `composer test:pest:full` (no TIA) | 5,181 passed, 2 skipped, 24,893 assertions |
| After `71bc3faa1` | both SystemAdmin test files 11/11, `tests/Arch` 74/74, Rector, PHPStan, type coverage 100%, lint |
| CI on `727fcc068` | all shards, browser tests, audits green |

### Mutation audit (12 production hunks, worst-first)

Each hunk was reverted alone and its narrowest tests were required to fail.

- 11 of 12 failed on the first run: `User::memberOf()` pivot branch, `sentBy()->typed()`, the store's `messageAttributes()` origin write, `visibleTo()->withoutSynthetic()`, the search filter, the typed supersede anchor, the opener prompt, the failed-turn origin insert, `hasUserChatMessage`, the `<turn>` block, and the title gate.
- **Gap closed:** dropping `whereRelation('conversation', 'workspace_id', ...)` from `AgentConversationMessage::ownedBy()` left all 1,361 Chat and Onboarding tests green. The old `TranscriptScope` join had the same hole. Added `lists no messages from a conversation in another workspace of the same user` to `ConversationWorkspaceScopingTest`; it now fails on that revert.

### Migration rehearsal on the production clone

Both migrations ran against a copy of the nightly pseudonymised prod clone (2,306 messages): 28ms total. 67 `continuation` rows plus 3 `[approval]` echoes became `resume`, the one legacy greeting became `greeting`. Afterwards 0 rows keep `meta.kind`, and 0 typed rows still hold a legacy prompt. The three assistant rows that start with `[approval]` were left alone, as intended. The three approval prefixes match every writer in git history (#209, #326, #345).

### Walked end to end (real Anthropic turns, Horizon on redis, Reverb)

Horizon was restarted from this checkout before the walk. No `local.ERROR` or `WARNING` was logged during it.

**J1. Typed request, approve, resume turn.** Stored rows: `user/typed`, `assistant`, `user/resume` holding only "The user decided the proposals above.", `assistant` "Created FP803 Origin Test Co". After reload the transcript shows no synthetic bubble.

transcript after resume

Searching all chats for the stored opener finds nothing, while searching for "FP803" finds the conversation. In-conversation search for the opener returns `{"matches": []}`.

search for opener

**J2. Reject, then double approve.** Discard stores a `resume` row and the reply "You rejected the new company, so nothing was created." Clicking Create twice in the same tick produced one company, one `resume` row and one reply.

rejected proposal

**J3. Setup greeting for a new user.** The signup listener created the empty setup thread. Opening it streamed a greeting that follows the `<turn>` directive (first name, first and last stage, ask for data). The stored row is `user/greeting` with the short opener. The activation checklist stays at 0/4 with "Ask Rela about your pipeline" open, while the owner who typed went from 1/4 to 3/4.

setup greeting

checklist after greeting

**J4. SystemAdmin.** The message list shows the Origin badge: Greeting and Resume on synthetic user rows. Before the fix, assistant rows read "Typed".

sysadmin origin column before the fix

After `71bc3faa1` the badge shows only on user rows; the assistant row and its view page show the empty placeholder.

sysadmin origin on user rows only

**J5. Reload while a typed turn streams.** 1.5s after send, a reload still rendered the typed bubble while the newest stored rows were the previous turn: `TurnPresence` v2 carries `origin` correctly. A resume turn in flight renders no bubble.

**J6. API member scoping with a real token.** `POST /api/v1/tasks` with the owner plus a pivot member as assignees returned 201. A user from another workspace returned 422 as an assignee and as `account_owner_id` on `POST /api/v1/companies`. A member as account owner returned 201.

All seeded users, workspaces, companies, tasks, tokens and conversations were deleted afterwards (counts verified 0). The rehearsal database was dropped.

### Findings

**1. Medium, accepted by the author. A `ProcessChatMessage` queued before the deploy crashes after it.** Reproduced by unserializing a payload that carries `isContinuation` instead of `origin`:

```
Error: Typed property Relaticle\Chat\Jobs\ProcessChatMessage::$origin must not be accessed before initialization
```

It hits every in-flight message, typed ones too, at `ProcessChatMessage.php:170`. `failed()` then throws again in `opensTheThread()`, so `TurnPresence::clear()` and `ChatStreamFailed` never run and the spinner stays until the presence TTL. The credit refund does run first. The window is the few seconds between code switch and worker restart, plus any job sitting in a `release()` backoff. Self-hosted upgrades (`docker compose pull && up -d`) have no drain step. #735 shipped the same shape when it added `$attachment`.

**2. Low, fixed in `71bc3faa1`. Assistant rows showed origin "Typed" in SystemAdmin.** The column default `typed` applies to every role (`AgentConversationMessageResource.php:84`, `MessagesRelationManager.php:41`, the infolist entry at `:53`). No query reads origin on assistant rows, so this was display only. `AgentConversationMessage::userOrigin()` now owns "the origin of a user row", and the list column, relation manager column and infolist entry read it with the empty placeholder. Covered by `shows an origin only on user rows` and `shows an origin only on user rows in the messages relation manager`; both fail when `userOrigin()` returns the raw origin.

**3. Low, report. Old workers can write a legacy continuation after the backfill.** Between `migrate` and the Horizon restart, an old worker still stores `meta.kind = continuation` with the column default `typed`, and the full prompt would render as a user bubble. The backfill never runs again. Same window as finding 1, so the same decision (accepted) covers it.

**4. Info. Hidden context reaches jobs dispatched mid-turn.** `Context::scope()` restores in a `finally` (verified in `vendor`), but `SuggestNextSteps`, the failover re-dispatch and queued broadcasts carry `chat.message_origin`. None stores a user message today, and the failover passes its own origin. Laravel flushes context per job.

**5. Info, follow-up outside this diff.** `ChatController` still checks conversation ownership in PHP at lines 128, 205, 416, 456 and 531. Those checks allow a `null` `workspace_id`, which `AgentConversation::ownedBy()` does not, so moving them to the scope changes behavior and needs its own PR.

**Scope note.** `727fcc068` (LoginBrowserTest waits for the one-time-code input to boot) is unrelated to message origin and not mentioned in the PR body.

**Cleanup pass (reuse, simplification, efficiency, altitude):** no change applied. EXPLAIN on the prod clone shows the new `whereRelation` shapes run as semi-joins through `conversation_index` and the conversations primary key (transcript page 0.36ms, `hasUserChatMessage` 0.06ms), so no index is needed. The remaining suggestions were already there before this PR or save one line. **net: -1 lines possible** (the repeated member-id pluck in `TenantFkValidator`, which predates this PR).

## Decisions

1. **In-flight jobs across the deploy:** accept the window, as #735 did (option C). No shim.
2. **Origin badge:** user rows only (option A), applied in `71bc3faa1`.

---

_Assessed: correctness, over-engineering, conventions, project guidelines, appsec, functional behaviour (small-N browser), migration on prod-shaped data, query plans on the prod clone, deploy-window job compatibility._
_Not assessed: load and scale beyond the 2.3k-row prod clone, abuse and rate limiting, lock behaviour under concurrent chat writes during the migration. Review separately for a high-traffic surface._

fp-sha:71bc3faa1

PR #827, opened 2026-09-22
fix(chat): report rejections truthfully and show thinking after a decision

ManukMinasyan:
Three of five production resumes after a rejected proposal reported the write as done ("All sample records have been removed"), because the resumed turn's user message was content-free and the outcome lived only in the `<resolved_actions>` system block, whose header told the model to report every just-decided entry as completed. The resume opener now names each decision with its consequence (`REJECTED (nothing was written): delete sample_data "All sample records"`), `ResolvedActionText` owns that wording for both the opener and the block, and the prompt keeps one status-branched Resuming section instead of four copies. Real-model replays went from 5 of 6 wrong to 11 of 11 correct with no regression on approvals. The "Thinking…" indicator now shows between a decision and the first token: the shimmer keys on whether any live assistant bubble has text rather than on the last transcript row, which for a resumed turn is the assistant's own proposal reply. The chat rule and guideline paragraphs describing the resume row were updated and the agent guideline files recompiled.


ManukMinasyan:
Superseded by #829 (head branch renamed to `fix/chat-report-rejections-truthfully`).

PR #828, opened 2026-09-22
feat: capability-based roles, chat field settings, and import audit entries

ManukMinasyan:
Every authorization check now reads one capability map owned by `WorkspaceRole`, and an arch test fails any role-string check outside it. The Editor role becomes Member, with a migration for stored `editor` keys; Viewers lose export, Admins gain custom field management, and invites show a hint per role plus a Compare roles matrix that the help page mirrors. The assistant refuses writes the user's role cannot approve, and can change any custom field setting the panel exposes (such as a currency's decimal places) through the usual approval card. The activity log now records custom field definition changes, the custom field values an import updates, and a company's account owner. Each import appears on the workspace Activity page as one entry, marked failed when it fails, while record timelines keep their own rows tagged with the import file.


ManukMinasyan:
Superseded by #830 (branch renamed).

PR #833, opened 2026-09-23
fix(mail): stop digests to inactive users and park inactive recipients

ManukMinasyan:
## Why

The daily task digest went to every assignee forever, including users who never returned after signup and only hold the four seeded demo tasks. Daily digest sends grew from 4 to 138 a day between 08-24 and 09-11, and 120 of 159 recipients had not logged in for a week or ever. Gmail keys reputation on the DKIM domain that invitations share, so this stream is one of the drains behind invites landing in spam.

Postmark answers a hard-bounced or complaining address with HTTP 406, and the queued mail retried it every day (22 `failed_jobs` rows in the last week, all the same three addresses).

## What

- `notifications:send-task-digest` skips users with no login in the last 30 days (`last_login_at` is stamped on every `Login` event, including remember-cookie recalls).
- `DigestService` leaves system-created tasks (`creation_source = system`) out. A seeded task the user later edited is still excluded; custom-field edits do not bump `tasks.updated_at`, so "untouched" is not cheaply knowable and the edge case is rare.
- Mailables using `ParksBouncedRecipients` (`TaskDigestMail`, `ProTrialEndingSoonMail`) stamp `users.email_bounced_at` when the transport reports 406. A `MessageSending` listener then drops parked recipients of those mailables before the transport. Essential mail (sign-in codes, password resets, invitations) is untouched, so a parked user can still get in.
- An email change clears the flag.

## Tests

`tests/Feature/Email/BouncedRecipientsTest.php` drives the 406 through the real queued-mailable path with a throwing transport; `SendTaskDigestCommandTest` covers the two gates. All four new behaviours fail without the production change.

## Not in this PR

Marketing consent and the in-app opt-out are a separate PR. Turning off open and link tracking on transactional mail, and the sending-domain split, are Mailcoach and Postmark dashboard settings.


ManukMinasyan:
## Final review: stop digests to inactive users and park inactive recipients

**Verdict: `needs-human`**. Reason: 1 open question awaiting a human (below). There are no blocking findings, and CI is green on `faf35f58c`.

Reviewed head: `faf35f58c`. Since the original fix `94531b4d0`, the branch gained:
- `9a980a966`: my test change (see Fixes).
- `27dd7ffd3`: an "Email Bounced" column and filter on the sysadmin users table, with 2 tests.
- `5a386567a`: a merge of `main`.
- `faf35f58c`: a Pint fix for `tests/Browser/Chat/ProposalFieldEditTest.php`. #836 merged that file into `main` with a lint failure, which failed Code Quality on `main` and on this PR.

The change does what it says, and production data backs both halves. I found no blocking bug. One test gap from the mutation audit is now closed. The open question below is an operability choice, not a defect.

### What production says (read-only queries, 2026-09-23)

- `failed_jobs`, last 30 days: **69 `TaskDigestMail` failures** raising `HttpTransportException ... (code 406)` for 3 recipients, repeating daily. There were also **2 `ProTrialEndingSoonMail` 406s**. The 406 is the real HTTP status: `MailcoachApiTransport::doSendApi()` throws `HttpTransportException("... (code {$statusCode})")` for any status other than 200/204/403/422. The trait's `getInfo('http_code') === 406` check matches that contract.
- The same table has **4 `TaskDigestMail` failures from an expired Mailcoach TLS certificate** (`Symfony\Component\HttpClient\Exception\TransportException`). This is the non-HTTP path, and it must never park anyone. It had no test (see Fixes).
- Who the 30-day login filter drops, counting users assigned open non-system tasks: 20 logged in within 30 days (the only paying assignee is among them) and keep their digest. 19 are stale (30 days or more) and 43 never logged in. None of the dropped users is on a paying workspace.
- API or MCP users never fire `Login`, so they carry no fresh `last_login_at`. Two token users and one OAuth user are active without a recent login. **None of them holds a real task today**, so nobody loses a digest they would get.
- `tasks.creation_source` is `NOT NULL` (`2024_09_11_114549_create_tasks_table`), so `!= 'system'` has no NULL trap. In production, 12,131 of 13,168 live tasks are `system` demo rows.

### Fixes applied

| Commit | Change | Why |
|---|---|---|
| `9a980a966` | `BouncedRecipientsTest` now drives the **real `MailcoachApiTransport`** through Symfony's `MockHttpClient` instead of a hand-rolled failing transport. Adds `keeps the recipient when the provider cannot be reached`. | The mutation audit left the `instanceof HttpTransportException` guard green: deleting it passed every test. Production hit that path on 2026-08-26 (TLS expiry). The new test goes red without the guard. Using the vendor transport also pins the 406-to-`HttpTransportException` contract to the actual Mailcoach code, not a copy of it. |

`/simplify` ran over the branch diff with 4 reviewers (reuse, simplification, efficiency, altitude). **Lean already.** I applied nothing, for these reasons:
- The suggested `SubscriberTagEnum::applyRecencyWindow(Active30d)` has different semantics: it is a 7 to 30 day band.
- `whereNull('email_bounced_at')` in the command would make a second owner of the parking rule.
- A code comment on the observer would break the comment policy.
- A `Parkable` interface would still need the trait, because the trait carries `failed()`.

### Findings (none blocking)

1. **Medium, operability. Parking is permanent and silent.** `app/Mail/Concerns/ParksBouncedRecipients.php:33`
   - Only an email change clears `email_bounced_at` (`app/Observers/UserObserver.php:12`). A Postmark reactivation never flows back.
   - `27dd7ffd3` now shows the flag in the sysadmin users table, with a filter. There is still no action to clear it.
   - After the one `failed_jobs` row, every later drop in `DropBouncedRecipientsListener` returns `false` with no log. On-call cannot tell that a user stopped getting mail. See Open question 1.
2. **Low, correctness boundary. API/MCP-only users are excluded from the digest.** `app/Console/Commands/SendTaskDigestCommand.php:31`: Sanctum and Passport requests never dispatch `Login`, so these users keep a stale `last_login_at`. Zero are affected today (see production data above). Worth a line in the PR body so it is a known limit, not a surprise.
3. **Low, pre-existing, repo-wide. Every listener in `AppServiceProvider::boot()` fires twice.** Laravel's default event discovery registers `app/Listeners` (`php artisan event:list` shows `Listener` and `Listener@handle` for `Login`, `Verified`, `MessageSending`). The explicit `Event::listen` lines register them again.
   - For this PR's listener the only cost is one extra `users` query per digest. The mutation audit confirms the `Event::listen(MessageSending::class, ...)` line is redundant: removing it keeps every test green.
   - `NewSubscriberListener` on `Verified` running twice deserves a separate look.
   - This dates from #217. I left it out of this PR.
4. **Low, data at scale.** Production `users` has no index on `email`. The unique index that exists locally is missing in production (ULID-migration drift). The listener's `whereIn('email', ...)` therefore scans all 4.5k rows, once per parkable send (tens per day). Negligible now. Fold it into the missing-constraints cleanup.
5. **Low, guideline.** The 30-day `last_login_at` predicate is written inline next to the `atLocalHour` scope. The architecture guideline gives each model predicate a `#[Scope]` owner once it has readers. It has exactly one reader, so I kept it inline. `SubscriberTagEnum`'s 7 to 30 day band is a different fact.
6. **Low.** `failed()` parks every `To` address when one 406 comes back. Both parkable mailables have one recipient, so this is correct today. It becomes wrong if either mailable gains a second recipient.

### Open questions

1. **Log each park so on-call can see it?**
   - ☆ Add `Log::info('Parked inactive mail recipient', ['emails' => $emails])` to `failed()`. That is 1 line, and Flare/Nightwatch pick it up.
   - Leave it silent. The first 406 still lands in `failed_jobs`.
   - An "unpark" action next to the new sysadmin column, as a separate PR.

### Gate

| Check | Where | Result |
|---|---|---|
| Code Quality Checks (lint, rector, type coverage, phpstan) | CI on `faf35f58c` | pass |
| Tests, shards 1 to 5 | CI on `faf35f58c` | pass |
| Browser Tests | CI on `faf35f58c` | pass |
| `test:lint`, `test:refactor` (0 changed files), `test:type-coverage` (100%), `test:types` | local, merged tree | pass |
| Branch test files: bounce, digest command, trial mail, sysadmin users | local, `--no-tia` | 53 passed, 217 assertions |

The first local `test:pest:full` crashed in a worker: `git rev-parse` exceeded Pest's 5s timeout under machine load. That was an environment failure, and CI's full sharded run above is the canonical result.

**Mutation audit** (production hunks, each reverted alone, then the 3 affected test files run with `--no-tia`):

| Hunk | Revert result |
|---|---|
| `SendTaskDigestCommand` login filter | red: 2 tests (`skips users who have not logged in for 30 days`, `skips users who never logged in`) |
| `DigestService` system-task filter | red: `leaves seeded demo tasks out of the digest` |
| `ParksBouncedRecipients` on `TaskDigestMail` | red: 3 tests |
| `ParksBouncedRecipients` on `ProTrialEndingSoonMail` | red: `parks the owner when the trial notice bounces` |
| `failed()` 406 status check | red: `keeps the recipient on other provider errors` |
| `failed()` `instanceof` guard | **was green**; red after `9a980a966` (`keeps the recipient when the provider cannot be reached`) |
| `UserObserver::updating` reset | red: `delivers again once the user changes their email` |
| Observer `isDirty('email')` condition | red: `drops a parked recipient ...` |
| Listener: all-parked returns `false` | red |
| Listener: partial `To` rewrite | red: `still delivers parkable mail to the other recipients` |
| Listener: parkable-mailable guard | red: `still sends essential mail to a parked recipient` |
| Listener: `whereNotNull('email_bounced_at')` | red: 4 tests |
| `AppServiceProvider` `Event::listen` line | green, as expected: event discovery registers the listener as well (finding 3) |

`27dd7ffd3` (sysadmin column and filter) is covered by its own 2 tests and was not mutation-audited.

### End-to-end walk (local stack, real queue, real SMTP capture)

**Setup:** the dev database migrated with the new column; queue `redis`. No Horizon served this checkout, so I drained every job with this checkout's own `php artisan queue:work --stop-when-empty` (execution identity). Mail went to Herd's SMTP catcher on 127.0.0.1:2525.

**Fixtures:** three factory users per run in `America/Anchorage` (08:xx local at run time), tagged `fp-e2e-*`:
- active: never logged in
- stale: last login 45 days ago
- parked: logged in now, `email_bounced_at` set

Each workspace holds one web task due today, one overdue `system` task, and the 4 onboarding demo tasks the real seeder added. **All fixtures were deleted afterwards through `App\Actions\Jetstream\DeleteUser` (0 users and 0 tasks left).**

| # | Scenario | Result |
|---|---|---|
| J0 | Digest run before anyone logs in | Pass. `Queued 2`: only the two parked users. Both jobs `DONE`, **zero mail captured**. The stale and never-logged-in users were never queued. |
| J1 | Never-logged-in user signs in through the app login form, then the digest runs | Pass. The browser login stamped `last_login_at` (16:11:39). Herd row 903 went to that user, listing **only** "Call Dana about the renewal". The overdue `system` task is absent. |
| J2 | Parked user changes email in Settings > Profile. The flow: Confirm identity, then the verification mail, then the link opened in the browser. | Pass. "Email change requested" still reached the **parked** old address (row 904): essential mail is untouched. The new address got "Confirm your new email" (905). After the link, the toast read "Email address changed" and `email_bounced_at` was null. The next digest run delivered to the new address (row 907). The still-parked sibling stayed dropped. |

Digest for a freshly signed-in user
*Read-back: the box holds "DUE TODAY", "Call Dana about the renewal (active)" and "Due Sep 23, 2026". There is no Overdue section, so the overdue seeded task is absent.*

Digest delivered again after the email change
*Read-back: the box holds "Today's tasks, Parked", "Call Dana about the renewal (parked)" and "Due Sep 23, 2026". This is Herd row 907, sent to the new address.*

Videos of J1 and J2 are kept locally: `.context/finalize-pr/local/j1-login-then-digest.webm`, `j2-parked-email-change.webm`.

**Simulations, disclosed:**
- I backdated `last_login_at` and set `email_bounced_at` through the factory.
- I ran `notifications:send-task-digest` by hand rather than from the scheduler.
- I did not send a real 406 through the Mailcoach API. The live 406 contract rests on the production `failed_jobs` rows above plus the vendor-transport test.

### Not blocking, worth knowing before deploy

- The migration adds a nullable `timestamp` to `users` with no default and no backfill. That is a catalog-only change on Postgres, so there is no rewrite or lock risk. Self-hosted installs pick it up via `migrate`.
- On the first 08:00 run after deploy, the 3 addresses that fail daily each fail **one more time**, then get parked. `failed_jobs` stops growing for them after that.

_Assessed: correctness, over-engineering, conventions, project guidelines, appsec, functional behaviour (small-N browser plus real queue and SMTP capture), operability (406 contract against production failures, audience impact against production users, migration lock, query cost on the production `users` shape)._
_Not assessed: a live 406 round-trip through Mailcoach (no safe inactive address to send to), load beyond the production row counts above._

fp-sha:faf35f58c

